PR #5, opened 2019-03-21
Add x-unmock-client-user-agent header

ksaaskil:
- Add `X-Unmock-Client-User-Agent ` header to outgoing requests so we know where requests are coming from
- Could probably add much more data to the object but thought it would suffice to start from simple
- Should not require server changes as `*User-Agent*` is ignored by default (probably should be ignored in server-side always anyway?)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -61,3 +61,9 @@ export const endReporter = (
     }
   }
 };
+
+export const unmockUserAgentHeaders = () => {
+  return { lang: isNode ? "node" : "jsdom" };
+};
+
+export const UNMOCK_USER_AGENT_HEADER = "X-Unmock-Client-User-Agent";
--- a/src/util.ts
+++ b/src/util.ts
@@ -62,8 +62,8 @@ export const endReporter = (
   }
 };
 
-export const unmockUserAgentHeaders = () => {
+export const unmockUAHeaderValue = () => {
   return { lang: isNode ? "node" : "jsdom" };
 };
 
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -61,3 +61,9 @@ export const endReporter = (
     }
   }
 };
+
+export const unmockUAHeaderValue = () => {
+  return { lang: isNode ? "node" : "jsdom" };
```
idantene: Like you said - could be more specific. Even proliferation with e.g. `{ lang: "js", framework: isNode ? "node" : "jsdom", version: require("../package.json").version }` might be a good start (with `lang` reserved to the unmock client being used).

ksaaskil: I'll add the version, about the "lang" + framework I'm not that sure as it would just make analytics harder if one needs to split by both lang and framework? And framework may not make sense for other languages so wouldn't that be an extra complication?

idantene: Hm, that's true. Guess either one would be good enough for now :grinning:

ksaaskil: For reference, here's what Stripe does for Node: https://github.com/stripe/stripe-node/blob/master/lib/stripe.js#L13 and https://github.com/stripe/stripe-node/blob/master/lib/stripe.js#L294

idantene: Sure, has plenty of options. Their Python SDK adds a few more keys - https://github.com/stripe/stripe-python/blob/master/stripe/api_requestor.py#L238

ksaaskil: I'll just merge this now and ask Mike if there's a way to add the version that works in browser as well.

PR #6, opened 2019-03-21
[Feat] Updates the persistence layer to save metadata, and save response in one file

idantene:
- Saves response in `response.json` with headers under `headers` key and response body under `body` key.
- Saves an `unmock.metadata.yml` file that writes basic data about the original request and basic information about the client being used.
- Simplifies some of the read/write action in the persistence layers.
- Adds a `save: true` to tests for manual checking of returned mocks.

:arrow_right: 
- [x] Once approved, implement the same in `unmock-python`
- [ ] Future PR: simplify the `utils` library

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -85,7 +85,7 @@ const mHttp = (
       }
       // content being written or end of call!
       const resp = (res: IncomingMessage) => {
-        const protoOn = res.on  ;
+        const protoOn = res.on;
         res.on = (s: string, f: any) => {
           if (s === "data") {
             return protoOn.apply(res, [s, (d: any) => {
--- a/src/node.ts
+++ b/src/node.ts
@@ -27,19 +27,20 @@ export const initialize = (story: {story: string[]}, token: string | undefined, 
 const mHttp = (
   story: {story: string[]},
   token: string | undefined,
-  { logger, persistence, unmockHost, unmockPort, signature, save, ignore, whitelist }: IUnmockInternalOptions, cb: {
-    (
-        options: string | http.RequestOptions | URL,
-        callback?: ((res: http.IncomingMessage) => void) | undefined): http.ClientRequest;
-    (
-        url: string | URL,
-        options: http.RequestOptions,
-        callback?: ((res: http.IncomingMessage) => void) | undefined): http.ClientRequest;
+  { logger, persistence, unmockHost, unmockPort, signature, save, ignore, whitelist }: IUnmockInternalOptions,
+  cb: {
+    // tslint:disable-next-line:max-line-length
+    (options: string | http.RequestOptions | URL, callback?: ((res: http.IncomingMessage) => void) | undefined): http.ClientRequest;
+    // tslint:disable-next-line:max-line-length
+    (url: string | URL, options: http.RequestOptions, callback?: ((res: http.IncomingMessage) => void) | undefined): http.ClientRequest;
   }) => {
-  return (
-    first: RequestOptions | string | URL,
-    second: RequestOptions | ((res: IncomingMessage) => void) | undefined,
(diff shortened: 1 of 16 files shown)


src/node.ts
```diff
@@ -82,8 +83,9 @@ const mHttp = (
         // self call, we ignore
         selfcall = true;
       }
+      // content being written or end of call!
       const resp = (res: IncomingMessage) => {
-        const protoOn = res.on;
+        const protoOn = res.on  ;
```
ksaaskil: Extra space

src/persistence/fs-persistence.ts
```diff
@@ -1,48 +1,75 @@
...
 export default class FSPersistence implements IPersistence {
   private token: string | undefined;
+
+  constructor(private savePath = SAVE_PATH) {
+  }
+
+  public saveMetadata(hash: string, data: {[key: string]: string}) {
+    const target = this.outdir(hash, METADATA_FILE);
+    // First attempt to load existing data
+    const existingData = fs.existsSync(target) ? yml.safeLoad(fs.readFileSync(target, "utf-8")) : {};
```
ksaaskil: Is this meant to throw if the data is not proper yaml?

src/persistence/fs-persistence.ts
```diff
@@ -1,48 +1,75 @@
...
+    // Now add the new data as needed
+    for (const key of Object.keys(data)) {
+      existingData[key] = data[key];
```
ksaaskil: Is this somehow different than using the spread syntax like `existingData = { ...existingData, ...data }`?

src/util.ts
```diff
@@ -5,12 +5,23 @@ import isNode from "detect-node";
...
+}
+
 export const hostIsWhitelisted =
   (whitelist: string[] | undefined, host: string | undefined, hostname: string | undefined) =>
     whitelist &&
       ((host && whitelist.indexOf(host) !== -1)
         || (hostname && whitelist.indexOf(hostname) !== -1));
 
+// TODO: buildPath and endReporter have to be made easier and simpler to use...
+//        --> perhaps part of UnmockOptions as it is in unmock-python?
```
ksaaskil: I'd perhaps add these comments as issues or elsewhere rather than leave them clutter the code?

src/persistence/fs-persistence.ts
idantene: I figured it's better to throw on the rare cases where the data is malformed.
These are files that are not meant to be edited/changed, and are written by our system.

Do you feel it should continue silently?

src/persistence/fs-persistence.ts
idantene: Nope! I forgot about that, thanks! :)

idantene: Edit: Different because `existingData` is a `const`, so can't reassign to it, but otherwise it's identical.

src/util.ts
idantene: Added to JIRA

src/persistence/fs-persistence.ts
ksaaskil: Ok, just checking it's deliberate. Was just wondering if that data is something users are supposed to regularly edit and that would easily break. Probably better to throw.

src/persistence/fs-persistence.ts
ksaaskil: Perhaps it could be changed to `let? :) `Object.assign` would also be quite idiomatic js.

idantene: `const newData = {...existingData, ...data};` :wink:

idantene:
Woops, didn't squash :see_no_evil: 
My bad!

PR #7, opened 2019-03-24
Renames interface and tightens typing.

mikesol:


src/util.ts
```diff
@@ -5,12 +5,12 @@ import isNode from "detect-node";
...
 const querystring = isNode ? require("querystring") : require("querystring-browser");
 
-interface IOriginalRequestMetadata {
-  [key: string]: any;
-  requestHeaders: any;
-  requestHost: string;
-  requestMethod: string;
-  requestPath: string;
+export interface IPersistableData {
+  lang?: string;
```
idantene: Is the `[key: string]: any` not necessary anymore with this (i.e. does it automatically change `{ lang: "x", requestPath: "y" }` to `IPersistableData`?

idantene:
@mikesol should this be updated and merged then?

PR #8, opened 2019-04-15
Save metadata snapshots if run with Jest

idantene:
- If run with `jest`, create snapshots for each call, and save metadata as snapshot, e.g. after running `yarn test test/node.test.ts`, we get the following under `test/__snapshots__/node.test.ts.snap`:
```bash
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`unmock end to end node 1`] = `{
  "hash":"fd0c3e6f",
  "host":"www.behance.net",
  "method":"get",
  "path":"/v2/projects?api_key=u_n_m_o_c_k_200"
}`;

exports[`unmock end to end node 2`] = `{
  "hash":"e8276c8e",
  "host":"www.behance.net",
  "method":"get",
  "path":"/v2/projects/5630?api_key=u_n_m_o_c_k_200"
}`;

exports[`unmock end to end node 3`] = `{
  "hash":"d603ae6e",
  "host":"www.behance.net",
  "method":"get",
  "path":"/v2/projects/5630/comments?api_key=u_n_m_o_c_k_200"
}`;

exports[`unmock end to end node 4`] = `{
  "hash":"d01d13fe", 
  "host":"api.hubapi.com",
  "method":"post",
  "path":"/contacts/v1/contact/createOrUpdate/email/testingapis@hubspot.com/?hapikey=demo"
}`;
```

- The files are generated silently - the user doesn't need to add them; they may be checked into version control - do we want this?
- The idea is that after this, in the vscode plugin, we can match the test file to the matching snapshot (i.e. `node.test.ts` :arrow_right: `node.test.ts.snap`) and the matching test function by matching against `exports[METHOD N]`.
- The content of the matched exports will then be parsed (as it's JSON) and we can match:
  - **best case**: host + path + method are explicitly written down, so we can match against them (even one-to-many) and allow the quick transition to the unmock response.
  - **typical case**: the above are not explicit, but there are N calls for N exports, so we just allocate one per call
    - *caveat*: what about places where calls are dependent on conditionals? i.e. there are M > N calls in the code, but only N actually happen due to different routing logic?
  - **worst case**: we don't know how to allocate the calls; we present them all in one big bunch together in the form of `hash - GET host/path`.
- **Caveats**:
  - JS-only for the time being (which is fine, as the vscode-plugin is JS-only as well)
  - Jest only for now, can probably be extended to Mocha; we need to compile a list of supported frameworks though.

Would appreciate thoughts and such.


idantene:
Could also use e.g. https://github.com/bahmutov/snap-shot-it instead of jest!

ksaaskil:
Looks good to me 👍 Would perhaps put "snapshotters" somewhere else than `node.ts` to avoid polluting the already big file and to try to keep it focused on the core functionality.

idantene:
@mikesol 
@ksaaskil of course - done

Can I merge then? :innocent:
The big caveat is that it generates more file that the user might not want, and if they don't use the vscode plugin - don't need either.

idantene:
- Will test a pseudo-backend with Mocha as well, to see how it integrates
- A quick hack could be to just add a dependency on `expect` and/or `jest-snapshot`?
- Need to verify it works well with webpack

idantene:
Downside: tests with identical names will collide. Wondering what's the best course of action regarding this (although perhaps more suitable for a different PR).

PR #12, opened 2019-04-24
Defaults to use of local mocks

mikesol:


mikesol:
Saw that too, it is for lerna. As we don't ship with lerna we get the clean bill of health from snyk, but definitely something to keep an eye on!

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

packages/unmock-core/src/persistence/failing-persistence.ts
```diff
@@ -4,15 +4,37 @@ export default class FailingPersistence implements IPersistence {
...
   public loadAuth() {
     throw new Error("not implemented");
   }
+  public hasHash() {
+    throw new Error("not implemented");
+    return false;
+  }
+  public loadMeta() {
+    throw new Error("not implemented");
+    return {};
```
idantene: `return` after `throw`?

packages/unmock-node/package.json
```diff
@@ -16,6 +16,7 @@
     "follow-redirects": "^1.7.0",
     "ini": "^1.3.5",
     "js-yaml": "^3.13.1",
+    "mitm": "^1.7.0",
```
idantene: Woohoo!
Notice that this version does not match the one in the main `package.json` [here](https://github.com/unmock/unmock-js/blob/0afba86bc25abeb5c8cc5caafb862e70a0baaf0e/package.json)

packages/unmock-jsdom/src/persistence/local-storage-persistence.ts
```diff
@@ -1,29 +1,30 @@
-import * as yml from "js-yaml";
 import * as path from "path";
-import { IPersistableData, IPersistence } from "unmock-core";
+import { IMetaData, IPersistence, IRequestData, IResponseData } from "unmock-core";
 
 // Uses directory structure for keys in window local storage
-const UNMOCK_KEY = ".unmock";
+const UNMOCK_FILE = ".unmock";
```
idantene: `s/_FILE/_KEY/g`
This is still in browser, there are no files, are there?

packages/unmock-jsdom/src/persistence/local-storage-persistence.ts
```diff
@@ -48,4 +67,24 @@ export default class BrowserStoragePersistence implements IPersistence {
     return path.join(outdir, ...args);
   }
 
+  private genericSave<T>(hash: string, fn: string, data: T) {
```
idantene: Use of `T` generics seems like a leftover from `unmock-node`. Is there any reason `any` would not work here? Alternatively, shouldn't the calls in this `persistence` layer then use e.g. `this.genericLoad<IResponseData>(hash, RESPONSE_FILE)`?

packages/unmock-jsdom/src/backend/index.ts
```diff
@@ -117,6 +116,28 @@ export default class JSDomBackend implements IBackend {
           return XMLHttpRequestSetRequestHeader.apply(this, [name, value]);
         }
       };
+      const doEndReporting = (fromCache: boolean, responseBody: string, responseHeaders: any) =>
+        endReporter(
+          "HASH", // fix!
```
idantene: Was this meant to be fixed before the PR?

packages/unmock-node/src/__tests__/with-credentials.test.ts
```diff
@@ -30,4 +29,11 @@ test("credentials written to .unmock/credentials work just like a token", async
     "https://www.behance.net/v2/projects?api_key=u_n_m_o_c_k_200",
   );
   expect(typeof projects[0].id).toBe("number");
+  // test to make sure cache works
+  const {
+    data,
+  } = await axios(
+    "https://www.behance.net/v2/projects?api_key=u_n_m_o_c_k_200",
+  );
+  expect(typeof data.projects[0].id).toBe("number");
```
idantene: Is this a visual test (based on the logger output)?
If so, it might be better to provide a `MockLogger` class so we can check if it used "cached" or not :thinking:

packages/unmock-node/src/backend/index.ts
```diff
@@ -1,8 +1,10 @@
-import fr from "follow-redirects";
```
idantene: Happy to be rid of this ;) Should `follow-redirects` be removed from the dependencies then?

packages/unmock-node/src/backend/util.ts
```diff
@@ -0,0 +1,4 @@
+export const rawHeadersToHeaders = (rawHeaders: string[]) => new Array(rawHeaders.length / 2)
```
idantene: `headersArrayToObject`?

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
     token: string | undefined,
     options: IUnmockInternalOptions,
   ) {
-    fr.http.request = mHttp(story, token, options, httpreq);
-    http.request = mHttp(story, token, options, httpreqmod);
-    fr.https.request = mHttp(story, token, options, httpsreq);
-    https.request = mHttp(story, token, options, httpsreqmod);
+    mitm = Mitm();
+    mitm.on("connect", (socket: any, opts: any) => {
+      if (opts.shouldBypass) {
```
idantene: Wouldn't this be `isWhiteListed(socket......)`?
It would make `mHttp` also shorter.

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
-      ignore,
-      ro.method,
-      ro.path,
-      signature,
-      story.story,
-      unmockHost,
-      token !== undefined,
+  const { Host, ...rawHeaders } = rawHeadersToHeaders(req.rawHeaders);
+  const [h, p] = Host.split(":");
+  if (hostIsWhitelisted(whitelist, h, p)) {
```
idantene: `hostIsWhitelisted` is meant to accept `host` and `hostname` as 2nd and 3rd argument.
Perhaps it's time we retire from `host` vs `hostname` and just combine them `host || hostname`. Port is irrelevant for whitelisting.

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
-      ro.hostname,
-      ignore,
-      ro.method,
-      ro.path,
-      signature,
-      story.story,
-      unmockHost,
-      token !== undefined,
+  const { Host, ...rawHeaders } = rawHeadersToHeaders(req.rawHeaders);
+  const [h, p] = Host.split(":");
```
idantene: Could we do `[host, port]`? It's harder to keep track of single-letter variables later on.

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
+    {
+      ...(requestBody.length > 0 ? { body: requestBody.map((buffer) => buffer.toString()).join("")} : {}),
+      headers: rawHeaders,
+      host: h,
+      method: req.method,
+      path: req.url,
+    },
+    {
+      headers: responseHeaders,
+      ...(responseBody.length > 0 ? { body: responseBody.map((buffer) => buffer.toString()).join("")} : {}),
```
idantene:
```typescript
const BufferArrayToBodyObjectOrNull = (buf: Buffer[]) => buf.length > 0 ? { body: buf.map(buffer => buffer.toString()).join("") } : {}
```

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
+        });
+        res.write(response.body);
+        res.end();
+        doEndReporting(true)(hash, body, response.headers, !response.body ? [] : [Buffer.from(response.body)]);
+        return {
+          body,
+          hash,
+          headers,
+          host,
+          intercepted: true,
```
idantene: Is `intercepted` effectively `fromCache`?

packages/unmock-node/src/backend/index.ts
```diff
@@ -29,168 +32,159 @@ const mHttp = (
...
-    http.request = mHttp(story, token, options, httpreqmod);
-    fr.https.request = mHttp(story, token, options, httpsreq);
-    https.request = mHttp(story, token, options, httpsreqmod);
+    mitm = Mitm();
+    mitm.on("connect", (socket: any, opts: any) => {
+      if (opts.shouldBypass) {
+        socket.bypass();
+      }
+    });
+    mitm.on("request", async (req: IncomingMessage, res: ServerResponse) => {
```
idantene: why `async`?

packages/unmock-node/src/persistence/fs-persistence.ts
```diff
@@ -65,10 +83,29 @@ export default class FSPersistence implements IPersistence {
     return config.unmock.token;
   }
 
-  private outdir(hash: string, ...args: string[]) {
+  private outdir(hash: string, mkd: boolean, ...args: string[]) {
     const outdir = path.normalize(path.join(this.savePath, hash));
-    mkdirp.sync(outdir);
+    if (mkd) { mkdirp.sync(outdir); }
```
idantene: Nice catch!

jest.setup.js
```diff
@@ -1 +1,5 @@
-jest.setTimeout(30000);
+const mkdirp = require("mkdirp");
+const rimraf = require("rimraf");
```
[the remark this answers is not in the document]
mikesol: We want to start from a _tabla rasa_ with testing every time, so I think it's important we delete everything when we start.  The creation of `.unmock` is part of the normal flow.  I don't think we need rimraf, though - another way to do it is to make the testing folder random every time (ie `tmp/<uuid>`).  Can that be the subject of a different PR?

packages/unmock-core/src/backend/index.ts
```diff
@@ -2,6 +2,7 @@ import { IUnmockInternalOptions } from "../options";
 
 export interface IBackend {
   initialize: (
+    userId: string,
```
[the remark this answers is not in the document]
mikesol: We need `userId` for the hash function.

packages/unmock-core/src/index.ts
```diff
@@ -28,8 +31,10 @@ export const defaultOptions: IUnmockInternalOptions = {
   backend: new FailingBackend(),
   ignore: { headers: "\\w*User-Agent\\w*" },
   logger: new FailingLogger(),
+  mode: Mode.CALL_UNMOCK_FOR_NEW_MOCKS,
   persistence: new FailingPersistence(),
-  save: false,
+  save: true,
+  share: true,
```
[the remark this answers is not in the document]
mikesol: Will get rid of, that will be for another PR.

packages/unmock-node/package.json
mikesol: That version is for `@types`.  Also, I just noticed that `expect` is in the wrong `package.json` - it should be in one of the sub-projects. Will fix.

packages/unmock-jsdom/src/persistence/local-storage-persistence.ts
mikesol: No, but it was originally named `UNMOCK_FILE`, so I changed it back to that. The reason IIRC was to have the same naming convention everywhere.

packages/unmock-jsdom/src/persistence/local-storage-persistence.ts
mikesol: The generics are inferred, so using `<IResponseData>` is optional.

packages/unmock-jsdom/src/backend/index.ts
mikesol: Ah crap, yeah :-(

mikesol: The problem is in reporting to the CLI. Everything works, but it is printed as `"HASH"`, which is why no test caught it. I will write tests for `endReporter` at some point.

packages/unmock-node/src/__tests__/with-credentials.test.ts
mikesol: True.

packages/unmock-node/src/backend/index.ts
mikesol: Yup!

packages/unmock-node/src/backend/util.ts
mikesol: `rawHeaders` is a node convention that means "an array of headers".

packages/unmock-node/src/backend/index.ts
mikesol: Sorry, I don't understand this.  Could you explain?

packages/unmock-node/src/backend/index.ts
mikesol: Completely agree!

packages/unmock-node/src/backend/index.ts
mikesol: The problem is that there are other variables named host and port later on. Agreed that this sucks, though...

packages/unmock-node/src/backend/index.ts
mikesol: Good idea!

packages/unmock-node/src/backend/index.ts
mikesol: mmm... yes

packages/unmock-node/src/backend/index.ts
mikesol: Stupidity on my part, gone now :-)

packages/unmock-node/src/persistence/fs-persistence.ts
mikesol: Thanks!

jest.setup.js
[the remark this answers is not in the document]
idantene: Definitely for a different PR. A temp folder would indeed be an ideal solution IMO, and we can do that by passing an argument of where to save the mocks (instead of `.unmock/save`)

packages/unmock-core/src/persistence/failing-persistence.ts
idantene: Ah, thought it was weird it was suddenly needed where as `loadToken` is okay. It's caused by the interface (returning undefined or a valid response).

packages/unmock-jsdom/src/persistence/local-storage-persistence.ts
idantene: Then it goes the other way around - why are they explicit in `unmock-node`? :sweat_smile:

packages/unmock-node/src/backend/util.ts
idantene: Yup, isn't the purpose of this function to convert an array to an object though?
Alternatively, why not use `headers` instead of `rawHeaders`?

packages/unmock-node/src/backend/index.ts
idantene: Yup! Why do we need to add the `shouldBypass` when creating a request (in `passthrough.ts`)?
The request is made to the Unmock API - which is whitelisted by default - so I thought it would make sense to use e.g.:
```typescript
mitm.on("connect", (socket: any, opts: any) => {
  if (hostIsWhitelisted(whitelist, opts.host)) {
    socket.bypass();
  }
}
```
(assuming we update `hostIsWhitelisted` to only use 2 arguments).

Then we can also remove the matching part in `mHttp`.

PR #14, opened 2019-04-25
Cleanup and simplify

idantene:
- `unmock-core` with a flatter hierarchy (+ introduces `interfaces.ts` and `constants.ts`).
- `UnmockOptions` class with default parameters; provides the object with more heavy lifting duties.
- Fixes test to represent changes and check cached versions
- Moves `snapshot` and `serialize` to `unmock-node` (they don't make sense in `unmock-core` for now, they're only used in `node`).
- *Perk*: adds wildcard whitelist! (e.g. `{ whitelist: "*.amazon.com" }`)

Request for `unmock-cloud`: return access token + userid in exchange for a refresh token.
Future PR: cache userId, test on pure offline mode (no internet for `ping` operation, etc)

(Many files touched by linting)


packages/unmock-core/src/__tests__/hash/v0/index.test.ts
```diff
@@ -1,187 +1,204 @@
-import { defaultOptions, ignoreAuth } from "unmock-core";
+import { ignoreAuth } from "unmock-core";
 import computeHash, { makeHashable } from "../../../hash/v0";
 
 describe("Hash function", () => {
```
mikesol: I would recommend checking this file out and then working on it again after the `hash` merge, as the `hash` tests are all new in the `hash` branch.

packages/unmock-hash/src/hash/v0/index.ts
```diff
@@ -2,31 +2,27 @@ import objectHash from "object-hash";
...
 
-export interface IHashableWithIgnoreV0 {
-  body?: string | {};
-  headers?: { [key: string]: string };
-  hostname?: string;
-  method?: string;
-  path?: string;
-  story?: string[];
-  user_id?: string;
+export interface IHashableV0 {
```
mikesol: IMO, we don't want any fields in `IHashable` to be undefined except fields, like signature, that can be legitimately undefined.  For example, if I understand correctly, this would allow someone to pass an empty object à la `computeHash({})` and it to return something.  The ideal behavior would be ts not even compiling that code.

As a side note, IIRC, `foo: string | undefined` is the same as `foo?: string`.

idantene: I understand the logic behind that - although it would still produce a valid hash.
Hence the compromise; `foo: string | undefined` is not the same as `foo?: string` - the latter is an optional field, while the former is a mandatory field whose value might be undefined.
At worse, it would allow someone to pass:
```typescript
computeHash({
  body: undefined,
  headers: undefined,
  mostname: undefined,
  method: undefined,
  path: undefined,
  story: undefined,
  user_id: undefined,
});
```

The bonus for it is allowing one less interface which IMO clutters the namespace.

idantene: `Partial<T>` to the rescue!

PR #16, opened 2019-04-26
Makes unmock a proxy for unmock-cli

mikesol:
Uses the same convention as Jest.


packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
-              logger.log(`      method: ${snapshots[call].method}`);
-              logger.log(`      host: ${snapshots[call].host}`);
-              logger.log(`      path: ${snapshots[call].path}`);
+    });
+
+  program
+    .command("list")
+    .action(() => {
+      const logger = new WinstonLogger();
+      glob("__snapshots__/**/*.snap", (e, matches) => {
```
idantene: We actually haven't discussed snapshots structure in our meeting.
Are we good with `__snapshots__` folder in the main hierarchy? Should it be placed under `.unmock`?

For now, I'd say we could leave it as is; perhaps changing to a const for easier access and to avoid magic numbers.

packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
+        if (e) {
+          throw e;
+        }
+        matches.forEach((match) => {
+          logger.log(match.substring(0, match.length - 5).substring(14));
```
idantene: This feels like it can be better accomplished using `path` (to not depend on specific naming conventions and such): `path.basename(match, ".snap")`

packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
+          const snapshots = require(`${process.cwd()}/${match}`);
+          const tests = Object.keys(snapshots)
+            .map(name => ({[name.split(" ").slice(0, -1).join(" ")]: new Array<string>()}))
+            .reduce((a, b) => ({ ...a, ...b}), {});
+          Object.keys(snapshots)
+            .forEach(name => {
+              tests[name.split(" ").slice(0, -1).join(" ")].push(name);
```
idantene: The `.split(" ").slice(0, -1).join(" ")` operations seems to repeat itself - should this be either a separate function or alternatively, one can iterate over `Object.keys(snapshots)` once and just build `tests` in one go?

README.md
```diff
@@ -47,143 +24,24 @@ Unmock works by overriding Node's `http.request` and `https.request` functions i
 
 ## Install
 
+Any of the subpackages in this project can be installed via `npm` or `yarn`.
+
 ```sh
-$ npm install --save unmock
+$ npm install --save unmock unmock-node
```
idantene: `npm install --save unmock-<integration>` (a-la https://www.npmjs.com/package/apollo-server)

packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
-    const makeHeaders = (headers: string[]) =>
-      headers
-        .map(h => ({
-          [h.substring(0, h.indexOf(":")).trim()]: h
-            .substring(h.indexOf(":") + 1)
-            .trim(),
-        }))
-        .reduce((a, b) => ({ ...a, ...b }), {});
+  program
+    .command("curl <url>")
```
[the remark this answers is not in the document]
mikesol: I like the `curl` syntax because it is familiar, but of course we should use whatever shortcuts people find meaningful.

packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
+      collect,
+      [],
+    )
+    .option("-X", "--request [command]", "Specify request command to use")
+    .option(
+      "-S",
+      "--signature [signature]",
+      "Specify a signature for a request in tokenless mode",
+    )
+    .action(async (url, options) => {
```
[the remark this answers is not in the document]
mikesol: Good call!

packages/unmock-cli/src/cli.ts
```diff
@@ -5,100 +5,102 @@ import fs from "fs";
...
-    glob("__snapshots__/**/*.snap", (e, matches) => {
-      if (e) {
-        throw e;
+  program
+    .command("open <hash>")
+    .option("-e, --editor [editor]", "editor (defaults to the EDITOR env variable or, in the absence of this, vi)")
+    .action((hash, options) => {
+      const logger = new WinstonLogger();
+      const editor = options.editor || process.env.EDITOR || "vi";
+      const path = `${process.cwd()}/.unmock/save/${hash}/response.json`;
```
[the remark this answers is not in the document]
mikesol: Agreed, we should have a separate PR that makes most paths user-settable.

packages/unmock-cli/src/cli.ts
mikesol: :+1:

mikesol:
Fixed various odds and ends.

PR #18, opened 2019-04-26
Cache userid

idantene:
If access token hasn't changed, attempt to read `userid` from a cached file. If the file does not exist (or the access token changed) - ask for `userid` from cloud.


idantene:
For now I moved `getToken` and `getUserId` away from `UnmockOptions`.
This can either be injection dependent (as it is now) or more OOP-ish in nature; not sure which way we'd like to go at this point.

PR #19, opened 2019-04-29
New snapshots

idantene:
- Place snapshots in save path (based on save path in persistence layer).
- Use JSON format for snapshots (removes use of `require`, easier for webpack, etc).

*WIP*: Try to log which line makes the call for easier linking between call and hashes.


packages/unmock-node/src/persistence/snapshot.ts
```diff
@@ -39,31 +38,25 @@ expect.extend({
...
-      .join("\n");
+    const stringContents = JSON.stringify(contents, undefined, 2);
     fs.writeFileSync(snapFile, stringContents);
     return { pass: true };
   },
 });
 
-export const snapshot = (obj: any) => {
+export const snapshot = (obj: any, savePath?: string) => {
+  if (savePath === undefined) {
```
ksaaskil: This is a bit unexpected to me, to return immediately if an optional argument is not supplied? Usually optional arguments are auto-filled by the method if they're missing, if the user doesn't want anything to happen they don't call the method. But it's just a matter of style, no reason to change it!

idantene: Good catch, it's a leftover :)
There's no point in passing an undefined `savePath` of course.

idantene:
Could not figure out how to fetch currently running line (or the line that introduced the http request) from Jest cli runner.
Some traceback is available but it either halts at `processTicksAndRejections` (node internals), or is hidden by Node. Will attempt further in a different PR in the future.

PR #25, opened 2019-05-06
[Fix] UnmockOptions implements IUnmockOptions

idantene:
Also adds a basic composition test

### DIFF
--- a/packages/unmock-core/src/__tests__/composition.test.ts
+++ b/packages/unmock-core/src/__tests__/composition.test.ts
@@ -4,27 +4,14 @@ import {
   UnmockOptions,
 } from "..";
 
+import { FailingPersistence } from "../persistence";
+jest.mock("../persistence");
+
 const silentBackend = {
   initialize: jest.fn(),
   reset: jest.fn(),
 };
-const silentPersistence = {
-  getPath: jest.fn(),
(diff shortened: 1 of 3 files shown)


packages/unmock-core/src/__tests__/composition.test.ts
```diff
@@ -0,0 +1,57 @@
+import {
```
mikesol: Check out https://jestjs.io/docs/en/es6-class-mocks - it is a cleaner way to mock classes.

idantene: Not sure how these would work w/o a class; can't find a way to use an interface cleanly with `jest.mock` in a way that TS won't complain.

idantene: Or did you mean just using `jest.fn()` in all pseudo-functions?

mikesol:
Lemme dig in, one sec.

mikesol:
See the most recent commit.

idantene:
> See the most recent commit.

:man_facepalming: my bad.

idantene:
A weird error in CircleCI though (can't reproduce locally) -
`    ENOMEM: not enough memory, read`

PR #28, opened 2019-05-07
Feature/cli tests

mikesol:


packages/unmock-cli/src/commands/curl.ts
```diff
@@ -1,24 +1,29 @@
...
     ...(options.data ? { data: JSON.parse(options.data) } : {}),
-    ...(options.headers ? { headers: makeHeaders(options.header) } : {}),
+    ...(options.headers ? { headers: makeHeaders(options.headers) } : {}),
     method: options.request || "GET",
     url,
   });
+  return data;
+};
+
+export const curl = async (url: string, options: any) => {
```
idantene: I guess `curl` no longer needs to be `async`?

packages/unmock-cli/src/commands/list.ts
```diff
@@ -15,37 +16,44 @@ interface IStringMap {
...
-        );
-        Object.keys(tests).forEach((test) => {
-          logger.log(`  ${test}`);
-          tests[test].forEach((call) => {
-            logger.log(`    hash: ${snapshots[call].hash}`);
-            logger.log(`      method: ${snapshots[call].method}`);
-            logger.log(`      host: ${snapshots[call].host}`);
-            logger.log(`      path: ${snapshots[call].path}`);
+export const listInternal = (logger: ILogger): Promise<undefined> => {
+  return new Promise((resolve, reject) => {
```
idantene: Do we plan on chaining commands to this?

packages/unmock-cli/src/__tests__/commands/list.test.ts
```diff
@@ -0,0 +1,11 @@
+import { curlInternal } from "../../commands/curl";
+import { listInternal } from "../../commands/list";
+
+test("list works", async () => {
+  // this makes sure that some test has happened
+  await curlInternal("https://www.behance.net/v2/projects", {});
+  const log = jest.fn();
+  await listInternal({ log });
```
idantene: Wouldn't it be better to have `log` as some store, so that we can check e.g. that `method`, `host`, `path` etc are all present in the output?

packages/unmock-cli/src/commands/list.ts
idantene: And since I see you've used `JSON.stringify` in `curl` - should we also maintain that implementation here? (build an object with the list content and let JSON format it?)

PR #30, opened 2019-05-07
Adds actions

mikesol:
Makes the actions interface accessible and provides sensible defaults

### DIFF
--- a/packages/unmock-core/src/__tests__/util.test.ts
+++ b/packages/unmock-core/src/__tests__/util.test.ts
@@ -89,6 +89,11 @@ test("build correct path without ignore", () => {
   expect(searchParams.get("headers")).toBe(
     '{"Authorization":"Foo","User-Agent":"Bar"}',
   );
+  expect(JSON.parse(searchParams.get("actions") || "")).toEqual([
+    "deserialize-json-body",
+    "deserialize-x-www-form-urlencoded-body",
+    "make-header-keys-lowercase",
+  ]);
   expect(searchParams.get("path")).toBe("/v1/x");
   expect(searchParams.get("ignore")).toBe(null);
   expect(searchParams.get("signature")).toBe("my-signature");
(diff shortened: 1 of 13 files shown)


packages/unmock-core/src/__tests__/util.test.ts
```diff
@@ -89,6 +89,11 @@ test("build correct path without ignore", () => {
   expect(searchParams.get("headers")).toBe(
     '{"Authorization":"Foo","User-Agent":"Bar"}',
   );
+  expect(JSON.parse(searchParams.get("actions") || "")).toEqual([
```
idantene: Is the `|| ""` needed for `toEqual`? Can't we use `toBe` instead if so, just to make the test clearer?

packages/unmock-core/src/options.ts
```diff
@@ -11,6 +12,11 @@ export enum Mode {
 
 export class UnmockOptions implements IUnmockOptions {
   public save: boolean | string[] = true;
+  public actions: ActionsV0 = [
```
idantene: Typing wise, this is not as clean IMO as e.g. `ActionsV0 | ActionsV0[]` would be.

packages/unmock-hash/src/hash/v0/index.ts
```diff
@@ -181,7 +181,7 @@ type ActionsActionsV0 =
   | "make-header-keys-lowercase"
   | "deserialize-json-body"
   | "deserialize-x-www-form-urlencoded-body";
-type ActionsV0 = ActionsActionsV0 | ActionsActionsV0[];
+export type ActionsV0 = ActionsActionsV0 | ActionsActionsV0[];
```
idantene: Re the above comment above clarity of types, concretely I would suggest dropping `ActionsV0`, renaming `ActionsActionsV0` to `ActionsV0` and just being specific about it where needed (accepting either a list of it or a single argument, or perhaps forcing a list).

PR #36, opened 2019-05-18
Adds challenge to hash

mikesol:


packages/unmock-hash/src/hash/v0/index.ts
```diff
@@ -19,7 +19,7 @@ export interface IHashableV0 {
   signature?: string;
 }
 
-interface ISerializedHashableWithIgnoreV0 {
+export interface ISerializedHashableWithIgnoreV0 {
```
idantene: Why is it exported now?

packages/unmock-hash/src/hash/v0/index.ts
```diff
@@ -188,7 +188,17 @@ export default (
...
-) =>
-  objectHash(
-    applyActions(makeIgnorable(initialInput, ignore || []), action || []),
-  ).substring(0, TRUNCATE_HASH_AT);
+  challenge?: string,
+  onError?: (
+    initialInput: IHashableV0,
+    afterIgnore: Partial<IHashableV0>,
+    afterActions: ISerializedHashableWithIgnoreV0) => void,
+) => {
```
idantene: I think with the proliferation of arguments here, it might make sense to include them as a single object so one does not have to pass `undefined`, and enforce the use of argument names.

I'm guessing that `challenge` is meant to be e.g. the hash computed by the cloud? It would be helpful to add a  comment describing it.

mikesol:
Yup! Will do. Sorry for fast-tracking these, I need them to get the server healthy again. Hopefully that will be cleared up soon!

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #45, opened 2019-06-13
Matcher 0.1

ksaaskil:
- Add matcher in unmock-core
- Simple "first match" logic with regex, exact and undefined
- Add new `ISerializedRequest` and `ISerializedResponse` to `unmock-core` interfaces. There are things like `IRequestData` for that already, but I'd perhaps prefer starting afresh 😇 
- [x] Move matcher to unmock-core 
- [x] Clean extra yarn.lock and other stuff
- [x]  Regenerate package-lock.json, it's in some weird state where it contains all kinds of stuff that's already deleted


packages/unmock-core/src/interfaces.ts
```diff
@@ -66,3 +66,23 @@ export interface IUnmockOptions {
   useInProduction?: boolean;
   mode?: Mode;
 }
+
+export interface ISerializedRequest {
```
mikesol: Do we want to leave the possibility open for protocol?

packages/unmock-matcher/src/matcher.ts
```diff
@@ -0,0 +1,59 @@
...
+type RequestComparator = (
+  intercepted: ISerializedRequest,
+  mock: ISerializedRequest,
+) => boolean;
+
+const requestsMatch: RequestComparator = (
+  interceptedRequest,
+  mockRequest,
+): boolean => {
+  function strictEqualMatch(property: keyof ISerializedRequest): boolean {
```
mikesol: Not sure if this is a clean pattern - could you just define `strictEqualMatch` to take additional params you need from the scope?

packages/unmock-core/src/interfaces.ts
ksaaskil: Added protocol.

packages/unmock-matcher/src/matcher.ts
ksaaskil: Moved it out of the function, also added prelim support for regexp in the mock.

packages/unmock-core/src/matcher.ts
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  property: keyof (ISerializedRequest & IMockRequest);
+  interceptedRequest: ISerializedRequest;
+  mockRequest: IMockRequest;
+}): boolean {
+  const mockRequestProperty = mockRequest[property];
+  const interceptedRequestProperty = interceptedRequest[property];
+
+  if (mockRequestProperty === undefined) {
+    return true;
```
idantene: Why return true on undefined property? I'd expect false by nature (no match if property doesn't exist?)

packages/unmock-core/src/matcher.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  if (mockRequestProperty instanceof RegExp) {
```
idantene: Since we added RegExp, it could also be nice to add a glob-like pattern (with a simple `*` being replaced with `.*?` for a `RegExp` pattern)

packages/unmock-core/src/interfaces.ts
```diff
@@ -34,3 +34,27 @@ export interface IUnmockOptions {
...
+  protocol: "http" | "https";
+}
+
+export type IMockRequest = {
+  [P in keyof ISerializedRequest]?: ISerializedRequest[P] | RegExp;
+};
+
+export interface ISerializedResponse {
+  body?: string;
+  headers?: any;
```
idantene: `headers?: { [key: string]: any }` might be a bit better typing wise

packages/unmock-node/src/serialize/index.ts
```diff
@@ -1,10 +1,70 @@
...
+    throw new Error("No host");
+  }
+
+  const { method, url } = interceptedRequest;
+  if (!url) {
+    throw new Error("Missing url.");
+  }
+  if (!method) {
+    throw new Error("Missing method");
+  }
```
idantene: Could maybe put all of these checks in a `validateRequestOrThrow(interceptedRequest);` to simplify the function?

packages/unmock-core/src/matcher.ts
ksaaskil: Hmm, to me it seemed most natural that if e.g. protocol is not defined, then it matches both http and https? Otherwise, it would be a bit cumbersome to define matching for both http and https. And by that token, I thought it would be simplest if the same "match for undefined" was true for every property.

packages/unmock-core/src/matcher.ts
ksaaskil: Yes that's a good idea, but I think it's simplest if the matcher for now works with regex and any globs are handled before the matcher by, e.g., [glob-to-regexp](https://www.npmjs.com/package/glob-to-regexp).

packages/unmock-core/src/interfaces.ts
ksaaskil: Yes, I was lazily just copying the existing types.

packages/unmock-core/src/matcher.ts
idantene: I can see how that's true for protocol, but what about `body` or `headers`?

I'm fine  with "match for undefined", but I think it needs to be explicitly mentioned somewhere then.

packages/unmock-core/src/matcher.ts
idantene: Yes of course, it should be before the matcher. I'm not a fan of how `glob-to-regexp` does things though (I prefer lazy quantifiers over greedy ones in general - I think it captures the intent of `glob` better). We could write our own IMO in this case (future PR?)

ksaaskil:
Sorry for the flurry of commits, noticed that tests for the serialization are missing and added stuff based on comments.

PR #49, opened 2019-06-18
Handle headers in request and response

ksaaskil:
- Add headers to `ISerializedRequest`. In unmock 0.0, `host` header was dropped from the serialized request, but I don't think there's any reason to do that here
- Add headers to the returned response with `res.setHeader`
- Add separate interfaces for incoming and outgoing headers, as they have different types in Node.js: outgoing headers allow numbers, as `res.setHeader(k, v)` allows that, and I don't see why the serialized response would not allow a number.
- Also permit more types to headers: `string[]`, `undefined` as Node.js allows this, unmock 0.0 did a type-cast to `{ [key: string]: string }` but I think that should be avoided unless required somewhere later on

### DIFF
--- a/packages/unmock-node/src/serialize/index.ts
+++ b/packages/unmock-node/src/serialize/index.ts
@@ -32,8 +32,13 @@ class BodySerializer extends readable.Transform {
 
 function extractVars(
   interceptedRequest: http.IncomingMessage,
-): { method: string; host: string; path: string } {
-  const { host: hostWithPort } = interceptedRequest.headers;
+): {
+  method: string;
+  host: string;
+  path: string;
+  headers: { [key: string]: string };
+} {
+  const { host: hostWithPort, ...headers } = interceptedRequest.headers;
 
   if (!hostWithPort) {
     throw new Error("No host");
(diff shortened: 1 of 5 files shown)


packages/unmock-node/src/serialize/index.ts
```diff
@@ -32,8 +32,15 @@ class BodySerializer extends readable.Transform {
 
 function extractVars(
   interceptedRequest: http.IncomingMessage,
-): { method: string; host: string; path: string } {
-  const { host: hostWithPort } = interceptedRequest.headers;
+): {
```
idantene: General note (unrelated to this PR) - it feels weird to me that we sometimes use arrow notation with `const` and sometimes use the `function` declaration.
Should we just be consistent and use one of them?

ksaaskil: I use whichever reads clearest to me, I don't think there's really any difference. I think if anyone's bothered by it they can change it, but personally I don't think that's a "company policy" kind of thing...

idantene: There a few differences, but just because they're so similar, I feel like this is almost like a linting/formatting thing. Not a policy, just consistency across the codebase.

ksaaskil: Ok, for me readability is more important than consistency but I can try and avoid function declarations.

PR #50, opened 2019-06-18
[Initial] DSL and schema management

idantene:
1. Implement a basic generator accepting a serialized request and returning a generated mock.
2. Parses an OpenAPI verbose spec and sets a rudimentary example on how to parse new information with `x-unmock-size`.
3. Sets up a template for fetching service specification object from a serialized request (so we can still maintain `jsdom` if we decide to)


idantene:
For some reason not all child dependencies are installed in CI... Passes tests locally. Will look into it tomorrow.

idantene:
> For some reason not all child dependencies are installed in CI... Passes tests locally. Will look into it tomorrow.

Fixed -> main conclusion (not written anywhere so at least here) - only top-most package should have a _package-lock.json_. Subpackages should not. Not sure how that affects installing single packages later on.

packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  test("Generates array of specified-size on consecutive calls", () => {
+    const req: ISerializedRequest = {
+      host: "NA",
+      method: "get",
+      path: "/pets",
+      protocol: "https",
+    };
+    const generator = mockGeneratorFactory(getSpecFromRequest);
+    expect(generator(req).response.body.length).toBe(4);
+    expect(generator(req).response.body.length).toBe(4);
```
ksaaskil: Duplicate?

packages/unmock-core/src/generator.ts
```diff
@@ -0,0 +1,126 @@
...
+    throw new Error(
+      `Can't find response for '${method} ${path}' for ${spec.info.title}`,
+    );
+  }
+  // We return the entire object so we can resolve references
+  // to request parameters etc as needed
+  return pathSchema;
+};
+
+const templateFromResponse = (response: any): any => {
```
ksaaskil: Also here even a type alias would make it a bit easier to understand what's going in and out, I think?

packages/unmock-core/src/interfaces.ts
```diff
@@ -73,6 +73,19 @@ export interface IMock {
   response: ISerializedResponse;
 }
 
+export interface IResponseCreatorFactoryInput {
+  mockGenerator: () => IMock[];
```
ksaaskil: Is the mock generator used for anything anymore? Now I'm quite confused of all the curried definitions :)

packages/unmock-core/tslint.json
```diff
@@ -4,6 +4,8 @@
         "tslint:recommended"
     ],
     "jsRules": {},
-    "rules": {},
+    "rules": {
+        "arrow-parens": false
```
ksaaskil: The ts-lint of packages should probably be inherited from the root tslint.json if possible? Not a topic for this PR of course.

idantene: Ah, I just saw it was present in the `unmock-node` linting, so I copied from there. It was annoying that one would highlight it as error and the formatter would remove the fix.

ksaaskil: Yup, maybe `tslint-config-prettier` would need to be setup for all packages as well to avoid the conflicts.

packages/unmock-core/src/interfaces.ts
idantene: Ah, just in your original part of the code in `responseCreatorFactory`, where `opts: IResponseCreatorFactoryInput` is passed, to pass `opts.mockGenerator` to `httpRequestMatcherFactory`.

packages/unmock-core/src/__tests__/generator.test.ts
idantene: I'll add a comment - meant to ensure that the response length is consistent across generations.

PR #51, opened 2019-06-19
[Initial] State management

idantene:
- Exposes a `serviceStoreFactory` that receives a function that generates a mapping between _service name_ and its respective OpenAPI Schema. Allows us to decouple the storage from the state management.
- Ideally each storage is defined by a backend, which will expose a `states` object. Then a call to modify a state would be: `state.github({...})` (in a fluent manner).
- To specify an endpoint, one could use `state.github("/users", {...})`. Finally, one can also specify REST methods specifically with e.g. `state.github.get({...})` or a combination e.g. `state.github.get("/users", {...})` for fine tuned control.
- Fluent API is currently backed with e.g. `state.github.get(...).post(...).....otherService(...)`, so the user can specify the service once and specify different endpoints as needed. This works nicely, has the caveats of: 1) making the code a bit more complicated (maintaining `previousService`, etc), and eliminating HTTP rest methods from the pool of possible service names (so a service cannot be named `get`, `post`, etc, and _all_ is also used internally, but that's changeable).
- Plan is to have the status code set from within the state (with e.g. `$code: 200` in top most level) for each request.

Most importantly, this piece of PR does not DO anything just yet -- it's mostly doing fluency tests, shows hierarchy, etc. Will embed the rest of the logic in a following PR.


packages/unmock-core/src/service/interfaces.ts
```diff
@@ -0,0 +1,22 @@
+export type IOASMappingGenerator = () => { [key: string]: OASSchema };
```
ksaaskil: What's the key? Could be given an easier-to-read name I think.

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface IStateMapping {
+  [key: string]: any;
+}
+
+// Type aliases for brevity
+export type OASSchema = any;
```
ksaaskil: This does not use `openapi3-ts` because Mike's writing the types for that, is that the reason?

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -0,0 +1,22 @@
+export type IOASMappingGenerator = () => { [key: string]: OASSchema };
+
+export interface IStateMapping {
+  [key: string]: any;
```
ksaaskil: Mapping from what to what?

packages/unmock-core/src/service/service.ts
```diff
@@ -0,0 +1,63 @@
+import { filter as _filter } from "lodash";
+import { HTTPMethod, IStateMapping, OASSchema } from "./interfaces";
+
+/**
+ * Implements the state management for a service
+ */
+
+// TODO: Is there room here for xstate, in the future?
+
+export class Service {
```
ksaaskil: Should this implement an `IService` interface?

packages/unmock-core/src/service/service.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Maintains a state for service
+  // First level kv pairs: paths -> methods
+  // Second level kv pairs: methods -> status codes
+  // Third level kv pairs: status codes -> response template overrides
+  // Fourth and beyond: template-specific.
+  // @ts-ignore
+  private state: IStateMapping = {};
```
ksaaskil: Typing for this would be quite useful, I think...

packages/unmock-core/src/service/service.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(private oasSchema: OASSchema) {}
+
+  get schema(): OASSchema {
+    return this.oasSchema;
+  }
+
+  public updateState({
+    // @ts-ignore
```
ksaaskil: Why `ts-ignore`?

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -0,0 +1,144 @@
...
+import {
+  HTTPMethod,
+  IOASMappingGenerator,
+  isRESTMethod,
+  IStateMapping,
+} from "./interfaces";
+import { Service } from "./service";
+
+class ServiceStore {
+  private statesMapping: { [key: string]: Service } = {};
```
ksaaskil: What's the key?

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -0,0 +1,144 @@
...
+    return (endpoint = DEFAULT_ENDPOINT, state: IStateMapping) =>
+      obj(endpoint, state, method as HTTPMethod);
+  },
+};
+
+const StateHandler = (initialCall: boolean) => {
+  return {
+    get: (store: any, serviceName: any) => {
+      if (initialCall) {
+        (store as ServiceStore).newFluent();
```
ksaaskil: To me this does not feel like service store's responsibility, keeping track of the state of user calls and changing its internal state as a side-effect. It's fine for the service store to expose side-effects like `changeState` or similar, but I think this belongs to some other component's responsibilities ,to avoid service store becoming a god class...

packages/unmock-core/src/util.ts
```diff
@@ -12,3 +12,45 @@ export const doUsefulStuffWithRequestAndResponse = (
     `We will do amazing things with ${metaData} ${requestData} and ${responseData}.`,
   );
 };
+
+export const getAtLevel = (
+  nestedObj: any,
+  level: number,
+  filterCb?: (key: string, value: any) => boolean,
```
ksaaskil: To me this does not seem like a "callback" (as in call back when done) but more like a filtering function?

idantene:
The types are there for brevity as you suggested earlier, until we figure
out exactly how they should look like.
Exceptions are the state interface (it can be recursive and accept anything
pretty much). I'll give it better names though for the time being.

The @ts-ignore is there because we throw errors on unused parameters.
Preferred that over removing them for this PR.

Will also decouple the fluency from state store, I just wanted to push the
PR :)

packages/unmock-core/src/service/interfaces.ts
idantene: Pretty much - it's a place holder for now following your suggestion [here](https://github.com/unmock/unmock-js/pull/50#discussion_r295165222).

Same for the below comments.

packages/unmock-core/src/service/service.ts
idantene: Sure - I don't mind that, I don't see the specific need for it at this point. Do you reckon there will be other classes implementing `IService`?

packages/unmock-core/src/service/interfaces.ts
ksaaskil: It's fine to have a place-holder, just asking why use `any` if a type would be available :)

packages/unmock-core/src/service/service.ts
ksaaskil: No I don't, but I don't think the general principle "program against an interface, not an object" really requires that? But it can be a lone class for now.

packages/unmock-core/src/service/interfaces.ts
ksaaskil: And `[key: string]: any` is still different to me than `[serviceName: string]: Specification` even if both passed the same type tests.

packages/unmock-core/src/service/service.ts
idantene: Fair enough, will add as a `TODO` for the next PR.

packages/unmock-core/src/service/interfaces.ts
idantene: Ah! Fair enough. Will update those to reflect their connection with one another.

packages/unmock-core/src/service/serviceStore.ts
idantene: Yup, very true. Was done hastily, updating :+1:

packages/unmock-core/src/interfaces.ts
```diff
@@ -84,13 +84,4 @@ export type CreateResponse = (
 // Used to load a service specification from a serialized request
 // Returns an object (parsed from specification)
 export type RequestToSpec = (sreq: ISerializedRequest) => any;
```
idantene: Temporary - will be removed soon enough

PR #52, opened 2019-06-25
Loader for service defs from file system

ksaaskil:
- Add `IServiceDef` interface, acting as input to the service parser
- Add `FsServiceDefLoader`, for loading "service definitions" from the file system for the service parser
- Does not create directories: I'd see some kind of facade like `FsServiceManager` using `FsServiceDefLoader` and `FsServiceDefCreator` with a unified interface, with separate classes for reading and creation tasks
- Some uncertainty in how to define the "base" directory: should it be without or including `__unmock__`. By default resolves to `process.cwd()/__unmock__` (if no configuration or environment variable set, I'd assume these are resolved somewhere higher up).


packages/unmock-core/src/interfaces.ts
```diff
@@ -85,3 +85,39 @@ export type CreateResponse = (
...
+  /**
+   * Contents of the service definition file
+   */
+  contents: string | Buffer;
+}
+
+/**
+ * Input to the service parser. Contains, e.g., the directory name and all available files.
+ */
+export interface IServiceDef {
```
idantene: The name here is a bit confusing. It's not a service definition at this point, but more of a service def collection?
Maybe `IServiceDefCollection` then?

packages/unmock-node/src/loaders/fs-service-def-loader.ts
```diff
@@ -0,0 +1,89 @@
...
+import path from "path";
+import { IServiceDef, IServiceDefLoader } from "unmock-core";
+
+const DEFAULT_SERVICE_SUBDIRECTORY = "__unmock__";
+
+export interface IFsServiceDefLoaderOptions {
+  servicesDir?: string;
+}
+
+const debugLog = debug("unmock:fs-service-def-loader");
```
idantene: This feels a bit weird as we haven't used `debug` at all so far in the project.
If we decide to use it, we should probably use it everywhere?

Or is this a leftover?

packages/unmock-node/src/loaders/fs-service-def-loader.ts
```diff
@@ -0,0 +1,89 @@
+import debug from "debug";
+import fs from "fs";
+import path from "path";
+import { IServiceDef, IServiceDefLoader } from "unmock-core";
+
+const DEFAULT_SERVICE_SUBDIRECTORY = "__unmock__";
+
+export interface IFsServiceDefLoaderOptions {
```
idantene: Could create an `interfaces.ts` and `constants.ts` in this folder as well, and move those there.

packages/unmock-node/src/loaders/fs-service-def-loader.ts
```diff
@@ -0,0 +1,89 @@
...
+  servicesDir?: string;
+}
+
+const debugLog = debug("unmock:fs-service-def-loader");
+
+/**
+ * Read services from file system. Base directory is either
+ * 1. Directory injected in configuration
+ * 2. Environment variable `UNMOCK_SERVICES_DIRECTORY`
+ * 3. `${process.cwd()}/__unmock__`
```
idantene: `process.cwd()` can be a bit tricky and might be a bit prone to errors if the user is navigating around the project workspace. I've used [app-root-path](https://www.npmjs.com/package/app-root-path). Not sure if that's the best solution, just a thought.

packages/unmock-node/src/loaders/fs-service-def-loader.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Resolve the absolute path to the directory where services live, with the flow:
+   * 1. Injected directory
+   * 2. Environment variable
+   * 3. ${process.cwd()}/__unmock__
+   */
+  private get servicesDirectory() {
+    const servicesDirectory = path.resolve(
```
idantene: This can probably be done when creating the object so the path doesn't need to be resolved every time `servicesDirectory` is called

packages/unmock-node/src/__tests__/loaders/fs-service-def-loader.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  it("throws for a non-existing directory", () => {
+    const serviceDefLoader = new FsServiceDefLoader({
+      servicesDir: "DEFINITELY_DOES_NOT_EXIST_I_HOPE",
+    });
+    try {
+      serviceDefLoader.loadSync();
+    } catch (e) {
+      expect(/does not exist/.test(e.message)).toBe(true);
+      return;
```
idantene: `expect(serviceDefLoader.loadSync).toThrow("does not exist");`

packages/unmock-core/src/interfaces.ts
ksaaskil: It contains everything that defines a single service (possibly with multiple files), to me `ServiceDefCollection` sounds like something that has to do with multiple services?

packages/unmock-node/src/loaders/fs-service-def-loader.ts
ksaaskil: I think we definitely need some logging and `debug` is the standard for that, I think?

packages/unmock-node/src/__tests__/loaders/fs-service-def-loader.test.ts
ksaaskil: True, I only had the async version earlier that worked better with the try-catch 👍

packages/unmock-node/src/loaders/fs-service-def-loader.ts
ksaaskil: 👍

packages/unmock-core/src/interfaces.ts
idantene: Ah, to me it read as if it's a collection of services. Alright then! :)

packages/unmock-node/src/loaders/fs-service-def-loader.ts
idantene: I don't doubt that - we just haven't used `debug` elsewhere yet.

packages/unmock-node/src/loaders/fs-service-def-loader.ts
ksaaskil: Hmm, they will then each contain only a single definition?

idantene: True, but if it's exported and used elsewhere it might be cleaner. If this are the only constants/interfaces you think will be used in `loaders` in general than I don't think it's needed. I thought there'll be more implementations coming, so it made sense (otherwise IMO it doesn't require its own folder)

packages/unmock-node/src/loaders/fs-service-def-loader.ts
ksaaskil: True, I'll leave a TODO note as we'll need to move the resolution of the directory out of the loader soon anyway.

PR #53, opened 2019-06-25
State paths with regex

idantene:
- Complements endpoints with matching regex pattern
  - at the moment this is very naïve - does not check against the provided `schema.type`, but rather accepts everything and provides a named regex matching for later retrieval.
- Moves method/endpoint checks to `Service` instead of `ServiceStore`
- Adds `IService` interface
- Adds tests
- Modifies tests for some edge cases (i.e. when there are no `paths` - it now throws an error; what does it mean to set a state for a service that cannot be accessed? perhaps we should even throw when trying to create such a service to begin with, considering it an invalid schema?)


packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -1,7 +1,8 @@
 import fs from "fs";
 import yaml from "js-yaml";
 import { mockGeneratorFactory } from "../generator";
-import { ISerializedRequest, OASSchema, RequestToSpec } from "../interfaces";
+import { ISerializedRequest, RequestToSpec } from "../interfaces";
```
idantene: When we'll get to a more generator-specific PR, these would probably change (i.e. `RequestToSpec` feels out of place at the moment)

packages/unmock-core/src/__tests__/serviceStore.test.ts
```diff
@@ -1,65 +1,135 @@
...
+    expect(store.petstore).toThrow("has no defined paths");
   });
 
-  test("state class with basic call to services", () => {
-    const state = serviceStoreFactory(EmptyPathsServicePopulator);
-    state.petstore(); // Should pass
+  test("Store with empty paths", () => {
+    const store = serviceStoreFactory(EmptyPathsServicePopulator);
+    expect(store.petstore).toThrow("has no defined paths");
+    expect(store.petstore.get).toThrow("has no defined paths");
```
idantene: These now throw because it doesn't make sense to set a state for a service that has no endpoints. Same for the changes below.

packages/unmock-core/src/__tests__/serviceStore.test.ts
```diff
@@ -1,65 +1,135 @@
...
   });
 
-  test("state class with REST method calls to services", () => {
-    const state = serviceStoreFactory(EmptyPathsServicePopulator);
-    state.petstore.post(); // Should pass
+  test("Store with non-empty paths with non-matching method", () => {
+    const store = serviceStoreFactory(PetsPathsServicePopulator);
+    expect(store.petstore.post).toThrow("Can't find any endpoints with method");
```
idantene: Same idea as above - what does it mean trying to set a state for "all paths with POST", where there are none? Hence it throws.

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -25,6 +25,18 @@ export interface IUnmockServiceState {
   [key: string]: UnmockServiceState;
 }
 
+export interface IService {
+  schema: OASSchema;
+  hasDefinedPaths: boolean;
+  findEndpoint(endpoint: string): string | undefined;
```
ksaaskil: What does this do?

packages/unmock-core/src/service/service.ts
```diff
@@ -1,37 +1,123 @@
...
-export class Service {
+
+export class Service implements IService {
   // Maintains a state for service
   // First level kv pairs: paths -> methods
   // Second level kv pairs: methods -> status codes
   // Third level kv pairs: status codes -> response template overrides
   // Fourth and beyond: template-specific.
-  // @ts-ignore
+  // @ts-ignore // ignored because it's currently only being read and not written
```
ksaaskil: Doesn't our TS config allow that :o Seems to me more like a thing linter would complain about.

packages/unmock-core/src/service/service.ts
```diff
@@ -48,17 +134,62 @@ export class Service {
...
-  //     // make sure the combination exists
-  //   }
-  // }
+  private __complementPathWithRegexp(newPath: string, oldPath: string) {
+    // Complements oldPath by adding an unmock extension for regex matching the path.
+    const newPathRegex = XRegExp(`^${newPath}$`, "g");
+    this.oasSchema.paths[oldPath][UNMOCK_PATH_REGEX_KW] = newPathRegex;
+  }
+
+  private __updateSchemaPaths() {
```
ksaaskil: Minor thing, but why use underscore if it's already a `private`? Seems like Pythonism to me :)

packages/unmock-core/src/util.ts
```diff
@@ -17,6 +17,7 @@ export const getAtLevel = (
   nestedObj: any,
   level: number,
   filterFn?: (key: string, value: any) => boolean,
+  noKeys: boolean = false,
```
ksaaskil: Brr, boolean argument :)

packages/unmock-core/src/service/service.ts
```diff
@@ -48,17 +134,62 @@ export class Service {
...
+    Object.keys(this.schema.paths).forEach((path: string) => {
```
ksaaskil: Quite a massive thing inside a `forEach`, could this be somehow split into smaller pieces or include a docstring?

packages/unmock-core/src/service/interfaces.ts
idantene: Will add documentation (_copy, paste_ your doc style in #52 )

packages/unmock-core/src/service/service.ts
idantene: It doesn't. I didn't want to take matters into my own hands as the state is temporarily not being used. Seems like a good protection IMO, linting-wise.

packages/unmock-core/src/service/service.ts
idantene: Very true :laughing: Renaming

packages/unmock-core/src/service/service.ts
idantene: I feel like it includes many docstrings, but I'll try and break it down further anyway. Thanks!

idantene:
> Looks good to me! I'm a bit worried of `Service` becoming a hard-to-understand state machine, it could be unavoidable but still we should do what we can to avoid it. IMHO it would be ideal if one could wrap all state changes to e.g. a `state` member and one could reset the "state" with a dead-simple `this.state = new ServiceState()` or `this.state.reset()`.

At the moment there's nothing that deals with `state` directly, it's only being written.
There's a lot of OAS management though, making sure requests are valid, paths are found, methods are okay, etc -- I'm not sure that's avoidable.

ksaaskil:
> > Looks good to me! I'm a bit worried of `Service` becoming a hard-to-understand state machine, it could be unavoidable but still we should do what we can to avoid it. IMHO it would be ideal if one could wrap all state changes to e.g. a `state` member and one could reset the "state" with a dead-simple `this.state = new ServiceState()` or `this.state.reset()`.
> 
> At the moment there's nothing that deals with `state` directly, it's only being written.
> There's a lot of OAS management though, making sure requests are valid, paths are found, methods are okay, etc -- I'm not sure that's avoidable.

Hmm, I don't understand how the OAS management requires `Service` to be changing its own state?

idantene:
> Hmm, I don't understand how the OAS management requires `Service` to be changing its own state?

You're mixing these up :)
The `Service` class verifies `X` against the OAS, and maintains a state. Maintaining the state is trivial.
The verification against OAS is the heavy lifting the `Service` class actually does.

IMO it makes no sense to call `states.myService({...})` if `myService` has no endpoints. Similarly, it makes no sense to set `states.petstore("/pet/2", {...})`, if the spec doesn't specify that endpoint.
The latter fails early instead of silently accepting an invalid state for an endpoint that does not exist (however, `states.petstore("/pets/2", {...})` is fine, as the spec defines `/pets/{petId}` and describes what `petId` actually is).

ksaaskil:
I agree 100%, I think we're speaking of different states here. I'm just saying `Service` should have as few changes to its internal members as possible.

idantene:
Ah, I agree with that too. The only change happening to internal members is supplementing every path with an `x-unmock-path-regex`, upon creating a Service. It's also a sanity check for OAS.

Hopefully, once we join `loas` in, we can make some simplifying assumptions (e.g. every path parameter is indeed described under `parameters`), thereby removing many conditionals.

PR #54, opened 2019-06-25
Service parser

ksaaskil:
- Add `ServiceParser` class to parse `IService` from `IServiceDef`. It only has one method and it does not take any constructor arguments so it does not really justify making it a class, but otoh I could see this becoming bigger with loas, support for json, etc.
- Add `name` to `IService`, imho it's an integral part of the interface?
- [x]  Conflicts with https://github.com/unmock/unmock-js/pull/55 so need to fix conflicts once that merged


idantene:
Merged #55 .
Before the PR -- if you want, we can modify `ServiceStore` still, so that it would accept `IService[]` instead of a mapping. The downside would be having to iterate over the array to find the matching service.
Plus side - one less type, and not having to pass something such as `{ petstore: new Service({schema ..., name: petstore }) }`.

ksaaskil:
> Merged #55 .
> Before the PR -- if you want, we can modify `ServiceStore` still, so that it would accept `IService[]` instead of a mapping. The downside would be having to iterate over the array to find the matching service.
> Plus side - one less type, and not having to pass something such as `{ petstore: new Service({schema ..., name: petstore }) }`.

I think `ServiceStore` could in its constructor create such a map from the list for quick access, imho that's its internal business as it's not calling anything external or injected there.

idantene:
Mind adding it in this PR then? Or is it another PR IYO?

packages/unmock-core/src/service/parser.ts
```diff
@@ -0,0 +1,34 @@
+import jsYaml from "js-yaml";
+import { IServiceDef, IServiceDefFile, IServiceParser } from "../interfaces";
+import { Service } from "./service";
+
+const PATTERN_FOR_KNOWN_FILENAMES = /^(?:index|spec|openapi)\.ya?ml$/;
+
+export class ServiceParser implements IServiceParser {
+  public parse(serviceDef: IServiceDef): Service {
+    const name = serviceDef.directoryName;
```
ksaaskil: This should probably be read from the yaml instead, or fallback to the directory name if not found? But then the spec is not valid openapi unless the name is prefixed with `x-`?

ksaaskil:
> Mind adding it in this PR then? Or is it another PR IYO?

Another imho to keep a scope in this...

idantene:
Alrighty, will wait for this to merge and again make the changes :innocent:

packages/unmock-core/src/__tests__/parser.test.ts
```diff
@@ -0,0 +1,43 @@
+import fs from "fs";
+import path from "path";
+import { IServiceDef } from "../interfaces";
+import { ServiceParser } from "../service/parser";
+
+const petStoreYamlString = fs
+  .readFileSync(path.join(__dirname, "__unmock__", "petstore", "spec.yaml"))
+  .toString("utf-8");
```
idantene: `readFileSync(..., "utf8")`

packages/unmock-core/src/service/parser.ts
idantene: Maybe just a `TODO` note here, we'll probably want to collectively decide on the heuristic here.
(Could be read from the yaml file, parsed from the `info.title` OAS, retrieved from the directory name, etc...)

packages/unmock-core/src/service/parser.ts
```diff
@@ -0,0 +1,34 @@
+import jsYaml from "js-yaml";
+import { IServiceDef, IServiceDefFile, IServiceParser } from "../interfaces";
+import { Service } from "./service";
+
+const PATTERN_FOR_KNOWN_FILENAMES = /^(?:index|spec|openapi)\.ya?ml$/;
```
idantene: Could move this to `constants`.
I would also add case insensitivity flag (so `OpenApi.yaml` will also match, for example).
IMO could also make it verbose with `new RegExp` (or `XRegExp`).

packages/unmock-core/src/service/parser.ts
```diff
@@ -0,0 +1,34 @@
...
+    const serviceFiles = serviceDef.serviceFiles;
+
+    const matchingFiles = serviceFiles.filter(
+      (serviceDefFile: IServiceDefFile) =>
+        PATTERN_FOR_KNOWN_FILENAMES.test(serviceDefFile.basename),
+    );
+
+    if (matchingFiles.length === 0) {
+      throw new Error(
+        `No idea what to do with: ${JSON.stringify(serviceFiles)}`,
```
idantene: Maybe a more indicative error message here? `Can't locate service specification file with pattern ... for ...` or something?

packages/unmock-core/src/service/parser.ts
```diff
@@ -0,0 +1,34 @@
...
+      );
+    }
+
+    const serviceFile = matchingFiles[0];
```
idantene: What do want to do if there is more than one matching file? :thinking:

packages/unmock-core/src/service/parser.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const contents: string =
+      serviceFile.contents instanceof Buffer
+        ? serviceFile.contents.toString("utf-8")
+        : serviceFile.contents;
```
idantene: Default encoding for `Buffer` is "utf8", this can be just `const contents = serviceFile.contents.toString();`.

ksaaskil: Yes, I just like being explicit!

packages/unmock-core/src/service/parser.ts
ksaaskil: No idea, I'd worry about that later.

packages/unmock-core/src/service/parser.ts
ksaaskil: Good point regarding the case insensitivity. I don't personally see any reason any other class would access this so I wouldn't put it to `constants.ts`, I'll just move it to be static class member.

PR #57, opened 2019-06-26
Add Matcher

ksaaskil:
- Add `Matcher` class taking in the OAS schema, possibly augmented with the regexes
- Create `Matcher` inside `Service` when the latter is created. It seems like it's easier to be detecting the matches "inside" the service instead of from the outside, but isolating that logic to a separate class seems useful
- Contains lots of overlap with the DSL-related matching that already exists in `Service`, but I wouldn't perhaps worry about duplicate code that much yet.
- Does not handle the case where servers object is inside the paths (or has variables)
- [x]  Get rid of `httpRequestMatcher`
- [x]  Add matching method in `ServiceStore`
- [x]  Add the matching method to `IService`


packages/unmock-core/src/service/matcher.ts
```diff
@@ -0,0 +1,132 @@
...
+      return directMatch;
+    }
+
+    const definedPaths = Object.keys(paths);
+    debugLog(`Searching for match for ${reqPath} for ${definedPaths}`);
+
+    for (const pathItemKey of Object.keys(paths)) {
+      const pathItemObject = paths[pathItemKey];
+      const pathRegex = pathItemObject[UNMOCK_PATH_REGEX_KW] as RegExp;
+      if (pathRegex === undefined) {
```
idantene: Can we assume it goes through `Service`? If so, then `UNMOCK_PATH_REGEX_KW` is promised to exist for each path.

packages/unmock-core/src/service/service.ts
```diff
@@ -44,13 +50,15 @@ export class Service implements IService {
     this.name = opts.name;
     // Update the paths in the first level to regex if needed
     if (this.schema === undefined || this.schema.paths === undefined) {
+      this.matcher = new OASMatcher({ schema: this.schema });
```
idantene: IIRC this doesn't copy `this.schema` into `OASMatcher`, but instead has the same object, right?
If that's the case, one could initialize `this.matcher` after `this.name` I think.

packages/unmock-core/src/service/service.ts
```diff
@@ -61,6 +69,10 @@ export class Service implements IService {
     return this.hasPaths;
   }
 
+  public match(sreq: ISerializedRequest): MatcherResponse {
```
idantene: Maybe `matchRequest` would be more verbose, but not really important :)

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -23,6 +25,8 @@ export interface IUnmockServiceState {
   [key: string]: UnmockServiceState;
 }
 
+export type MatcherResponse = any | undefined;
```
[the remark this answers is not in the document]
ksaaskil: Just a placeholder to remind that the matcher either returns an (hopefully-soon-to-be-typed) match or undefined.

packages/unmock-core/src/service/matcher.ts
```diff
@@ -0,0 +1,132 @@
...
+import { ISerializedRequest } from "../interfaces";
+
+import { MatcherResponse, OASSchema } from "./interfaces";
+
+const debugLog = debug("unmock:matcher");
+
+import { UNMOCK_PATH_REGEX_KW } from "./constants";
+
+// Just for readability until we have types
+type PathItemObject = any;
```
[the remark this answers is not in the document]
ksaaskil: Yep, `PathItemObject` would be [this](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#pathItemObject).

packages/unmock-core/src/service/matcher.ts
ksaaskil: Yes, initially I thought it might throw an error here for missing `UNMOCK_PATH_REGEX_KW` but then it also needs to be added in tests and it seemed simpler it would work without it as well...

packages/unmock-core/src/service/service.ts
ksaaskil: Yes that's true, personally I don't like the idea of sharing a mutable object and relying on it becoming updated at a later state but that would work in this case...

idantene: We can consider it a "one-time-mutable" though.
Alternatively, I can change it so that `Service` holds regex patterns elsewhere and just links to the original path, so the schema is indeed mutable.

Then one would use the `findEndpoint` method in `IService`.

ksaaskil: Hmm yeah, to me it would be clearest if a (static) call in the constructor would just copy `opts.schema` and add its own fields to the paths if they exist and then set the augmented version to `this.schema` and initialized the matcher with that. But it doesn't matter that much now, I'll just leave this for now if that's ok.

idantene: Of course.

If the aim is to keep the schema as is, I would rather avoid unnecessary copies and instead just create a `regexPathToPath` dictionary.

PR #58, opened 2019-06-26
Adds loas3

mikesol:


packages/unmock-core/package.json
```diff
@@ -17,6 +17,7 @@
     "axios": "^0.18.0",
     "js-yaml": "^3.13.1",
     "json-schema-faker": "^0.5.0-rc17",
+    "loas3": "0.0.3",
```
ksaaskil: Hard-coded version intentionally?

packages/unmock-core/src/service/parser.ts
```diff
@@ -28,7 +29,12 @@ export class ServiceParser implements IServiceParser {
         ? serviceFile.contents.toString("utf-8")
         : serviceFile.contents;
 
-    const schema = jsYaml.safeLoad(contents);
+    const { val: schema, errors } = loas3(jsYaml.safeLoad(contents));
```
ksaaskil: Might be easier for debugging to separate this into two steps as `safeLoad` can also throw an error, but for reasons not related to loas3. But not important now.

mikesol: Good idea. Also, I'm not sure if concatenating the messages with a newline is the best idea - it would be better to indicate which line numbers in the original document are responsible.

idantene: Sidenote for a future PR - we'll need to edit the schema passed to `safeLoad` to handle custom tags. We could separate this to two steps in that PR IMO.

PR #59, opened 2019-06-27
Types from loas3, cleanup in preparing for OASMatcher

idantene:
- Adds types from `loas3`
- Removes generator boilerplate in preparation for `OASMatcher` (#57 )
  - Removes generator test as well
- Updates types across `service`
- **Caveat**: removed `DEFAULT_REST_METHOD` as TypeScript complains when it's not explicit in some places.
- In many places `loas3` lacks a signature, making dynamic access harder.


packages/unmock-core/src/service/index.ts
```diff
@@ -1,16 +1,16 @@
...
-  IUnmockServiceState,
+  UnmockServiceState,
 } from "./interfaces";
 import { ServiceStore } from "./serviceStore";
 
 const saveStateProxy = (store: ServiceStore, serviceName: string) => (
   endpoint = DEFAULT_ENDPOINT,
-  method: HTTPMethod = DEFAULT_REST_METHOD,
-  state: IUnmockServiceState,
+  method: HTTPMethod = "all",
```
ksaaskil: Random comment: "all" isn't really an `HTTPMethod`, I guess? If that wasn't there, we could use this type in `ISerializedRequest.method` as well, I assume.

idantene: It's part of `HTTPMethod`, but not an actual HTTP method :roll_eyes: 
It's used to signal that any method should match (perhaps `any` is better than `all`?)

ksaaskil: Yes I understand its purpose for the state management, just wondering what I should call the type for `method` in `ISerializedRequest` if `HTTPMethod` already exists.

idantene: Hm, we could separate the state management then and just have a `type AugHTTPMethod = "all" | HTTPMethod`.

PR #60, opened 2019-06-27
Make Schema object immutable in Service

idantene:
- Removes `UNMOCK_PATH_REGEX_KW` and its insertion to schema `PathItem` object
- Use a mapping from endpoint (string) to regex to find matches
- Moves endpoint matching to `OASMatcher` (reduces code duplication and keeps scope) -> `Service` will now either delegate to `OASMatcher` or maintain & verify state, etc, abstracting `OASMatcher` from `ServiceStore`.
- Adds more types to `OASMatcher`.
- Adds `IOASMatcher` interface.

### DIFF
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -22,6 +22,10 @@ export interface IServiceMapping {
   [serviceName: string]: IService;
 }
 
+export interface IEndpointToRegexMapping {
+  [endpoint: string]: RegExp;
+}
+
 export interface IStateInput {
   method: HTTPMethod;
   endpoint: string;
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -28,10 +28,6 @@ export interface IServiceMapping {
   [serviceName: string]: IService;
 }
 
-export interface IEndpointToRegexMapping {
-  [endpoint: string]: RegExp;
-}
-
 export interface IStateInput {
   method: HTTPMethod;
   endpoint: string;
@@ -40,11 +36,6 @@ export interface IStateInput {
 export interface IServiceInput {
   schema: OpenAPIObject;
   name: string;
-}
-
-export interface IOASMatcher {
-  matchToOperationObject(sreq: ISerializedRequest): MatcherResponse;
-  findEndpoint(reqPath: string): string | undefined;
 }
 
 export type MatcherResponse = any | undefined;
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -1,7 +1,13 @@
 import { OpenAPIObject } from "loas3/dist/src/generated/full";
 import { ISerializedRequest } from "../interfaces";
 
-export { OpenAPIObject, Paths, Schema } from "loas3/dist/src/generated/full";
+export {
+  OpenAPIObject,
+  Paths,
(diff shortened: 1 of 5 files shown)


packages/unmock-core/src/service/service.ts
```diff
@@ -44,17 +46,21 @@ export class Service implements IService {
...
-    }
-    this.updateSchemaPaths();
     this.hasPaths = // Find this once, as schema is immutable
       this.schema !== undefined &&
       this.schema.paths !== undefined &&
       Object.keys(this.schema.paths).length > 0;
-    this.matcher = new OASMatcher({ schema: this.schema });
+
+    if (this.hasDefinedPaths) {
+      this.buildSchemaRegExps();
```
ksaaskil: I like the pure `this.endpointToRegexp = Service.buildSchemaRegExps(this.schema)` more than this pattern of filling in stuff as side-effect, but I guess that's a matter of taste.

idantene: That's fine, I'll update it as well.
Once I moved that it actually makes more sense to offload more of the matching to `OASMatcher`, so hold on for another re-PR please :)

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -32,6 +42,11 @@ export interface IServiceInput {
   name: string;
 }
 
+export interface IOASMatcher {
```
ksaaskil: Does this need an interface if it's only used internally by the `Service`?

packages/unmock-core/src/service/matcher.ts
```diff
@@ -30,10 +38,43 @@ export class OASMatcher {
...
+      }
+
+      const newPathRegex = XRegExp(`^${newPath}$`, "g");
+      mapping[path] = newPathRegex;
+    });
+    return mapping;
+  }
+
   private readonly schema: OpenAPIObject;
+  private readonly endpointToRegexMapping: IEndpointToRegexMapping = {};
```
ksaaskil: Removing initialization to `={}` here would ensure that it must be defined in the constructor in future as well.

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -22,6 +28,10 @@ export interface IServiceMapping {
   [serviceName: string]: IService;
 }
 
+export interface IEndpointToRegexMapping {
```
ksaaskil: Does this need an interface if it's only used internally by the matcher?

ksaaskil: I'd perhaps move it to `matcher.ts` and remove the `export`.

packages/unmock-core/src/service/interfaces.ts
idantene: Not really - was mostly thinking it could be useful if we'd like to extend it for other formats such as `RAML`, but it's definitely not needed now. I'll remove it then.

packages/unmock-core/src/service/matcher.ts
idantene: Nice catch, was a leftover from using `this.buildRegexpForPaths()` where typescript can't follow the logic to notice it's instantiated.

packages/unmock-core/src/service/interfaces.ts
idantene: Not really - it's just shorter than writing the mapping in several places and if we decide to change it for some reason, it takes effect in all relevant places.

I will move it to `matcher.ts` though.

PR #61, opened 2019-06-27
Bootstrapping for 0.1.0

ksaaskil:
- Add response creator factory, using `IServiceDefLoader` as the only dependency (for now)
- Change the mock generator to accept a template and return a serialized response, this will need adding the corresponding state when ready

### DIFF
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -31,18 +31,8 @@ export function responseCreatorFactory({
     parser.parse(serviceDef),
   );
   const serviceStore = new ServiceStore(services);
-  return (sreq: ISerializedRequest) => {
-    const responseTemplateOrUndefined = serviceStore.match(sreq);
-
-    if (responseTemplateOrUndefined === undefined) {
-      return undefined;
-    }
-
-    const responseTemplate = responseTemplateOrUndefined;
-
-    const sres = generateMockFromTemplate(responseTemplate);
-    return sres;
-  };
+  return (sreq: ISerializedRequest) =>
+    generateMockFromTemplate(serviceStore.match(sreq));
 }
 
 const setupJSFUnmockProperties = (_: UnmockServiceState) => {
@@ -59,8 +49,11 @@ const setupJSFUnmockProperties = (_: UnmockServiceState) => {
 };
 
 const generateMockFromTemplate = (
-  responseTemplate: Operation,
-): ISerializedResponse => {
+  responseTemplate: Operation | undefined,
+): ISerializedResponse | undefined => {
+  if (responseTemplate === undefined) {
+    return undefined;
+  }
   // 1. Take in state from DSL
   // TODO: Link with Matcher/Service
   const state = { $code: 200 };
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -1,6 +1,7 @@
 /**
  * Implements the logic for generating a response from a service file
  */
+import { Operation } from "loas3/dist/src/generated/full";
 import {
   CreateResponse,
   ISerializedRequest,
@@ -58,7 +59,7 @@ const setupJSFUnmockProperties = (_: UnmockServiceState) => {
 };
 
 const generateMockFromTemplate = (
-  responseTemplate: any,
+  responseTemplate: Operation,
 ): ISerializedResponse => {
   // 1. Take in state from DSL
   // TODO: Link with Matcher/Service
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -3,23 +3,46 @@
  */
 import {
   CreateResponse,
-  GeneratedMock,
-  IMock,
   ISerializedRequest,
-  RequestToSpec,
+  ISerializedResponse,
+  IServiceDef,
+  IServiceDefLoader,
 } from "./interfaces";
+
+import { IService } from "./service/interfaces";
+
 import { Schema, UnmockServiceState } from "./service/interfaces";
 
 // json-schema-faker doesn't have typed definitions?
 // @ts-ignore
 import jsf from "json-schema-faker";
+import { ServiceParser } from "./service/parser";
+import { ServiceStore } from "./service/serviceStore";
 
-export const responseCreatorFactory = (): CreateResponse => {
-  return (__: ISerializedRequest) => ({
-    body: "Nothing here yet",
-    statusCode: 200,
-  });
-};
+export function responseCreatorFactory({
+  serviceDefLoader,
+}: {
+  serviceDefLoader: IServiceDefLoader;
+}): CreateResponse {
+  const serviceDefs: IServiceDef[] = serviceDefLoader.loadSync();
+  const parser = new ServiceParser();
+  const services: IService[] = serviceDefs.map(serviceDef =>
+    parser.parse(serviceDef),
+  );
+  const serviceStore = new ServiceStore(services);
+  return (sreq: ISerializedRequest) => {
+    const responseTemplateOrUndefined = serviceStore.match(sreq);
+
+    if (responseTemplateOrUndefined === undefined) {
+      return undefined;
+    }
+
+    const responseTemplate = responseTemplateOrUndefined;
+
+    const sres = generateMockFromTemplate(responseTemplate);
+    return sres;
+  };
+}
 
 const setupJSFUnmockProperties = (_: UnmockServiceState) => {
   // TODO: use the state parameter to update values as needed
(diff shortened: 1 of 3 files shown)


packages/unmock-core/src/generator.ts
```diff
@@ -3,23 +3,46 @@
...
+    const responseTemplateOrUndefined = serviceStore.match(sreq);
+
+    if (responseTemplateOrUndefined === undefined) {
+      return undefined;
+    }
+
+    const responseTemplate = responseTemplateOrUndefined;
+
+    const sres = generateMockFromTemplate(responseTemplate);
+    return sres;
```
idantene: I'd assume either:
1. `generateMockFromTemplate` returns a valid response even for an empty template or undefined (i.e. empty body?)
2. Preferably, if the `ServiceStore` can't match, I'd expect an error rather than failing silently (following the "fail to success" paradigm).

packages/unmock-core/src/generator.ts
```diff
@@ -34,28 +57,16 @@ const setupJSFUnmockProperties = (_: UnmockServiceState) => {
...
-  // TODO: Link with matcher
-  const spec = getSpecFromRequest(sreq);
-  if (spec === undefined) {
-    throw new Error(`Can't find matching service for '${host}'`);
-  }
-  // 2. Use sreq to fetch the relevant path in the OAS spec
-  // const operation = getPathSchemaFromSpec(spec, method, path);
-  // 3. Fetch state from DSL
+const generateMockFromTemplate = (
+  responseTemplate: any,
```
idantene: `Response | undefined`?

packages/unmock-core/src/generator.ts
ksaaskil: Response creator is currently defined as returning undefined if it cannot return a response, and it's the backend's problem to deal with that (as it currently does by capturing the error and returning 501).

packages/unmock-core/src/generator.ts
ksaaskil: Why would it return undefined if given a template? If something goes wrong, it should throw an error I think.

idantene: No, the type for `responseTemplate` could be `Response | undefined` instead of `any`.

ksaaskil: What would empty template mean?

idantene: Nothing to generate. Empty response.

ksaaskil:
We're not failing silently anywhere here, error is returned if no match is found. And that's done in `unmock-node`, where `undefined` response is properly handled.

packages/unmock-core/src/generator.ts
ksaaskil: How's that expected in our case? Does OAS allow empty operation object?

packages/unmock-core/src/generator.ts
idantene: I'm not a big fan of either of those; ingesting a '501' and understanding it's an unmock-related error is not trivial at all.
I think it's a `core` issue, and the backends can catch the error if they so wish, but `ServiceStore` not being able to match a request is quite severe IMO

ksaaskil: Of course that's not how we want to handle the error, but that's again another PR to fix how to handle that.

packages/unmock-core/src/generator.ts
idantene: Technically, only the `responses` top level object is mandatory. All specific responses are optional (`responses: {}` is valid)

ksaaskil: But `operation: {}` is different than undefined?

idantene: True, but if we take the route of not throwing errors on failing to match (pt no. 1 in my upper comments), then `undefined` should probably be treated the same (or return `undefined`) and then:
```typescript
    const responseTemplateOrUndefined = serviceStore.match(sreq);

     if (responseTemplateOrUndefined === undefined) {
      return undefined;
    }

     const responseTemplate = responseTemplateOrUndefined;

     const sres = generateMockFromTemplate(responseTemplate);
    return sres;
```
becomes
```typescript
return generateMockFromTemplate(serviceStore.match(sreq));
```

Also, I'm not sure how `js-yaml` specs treat e.g. `responses: null`, that could be another case.

ksaaskil: Ok, that's just then pushing the undefined check one level deeper imho and could make it harder to debug. But I can change it of course if you think that's prettier.

idantene:
To me, returning `undefined` where it's an error is failing silently.
IMO it's better to have error handling with `try...catch` in the backend rather than checking for content, if one wishes to change the default behaviour of errors.

packages/unmock-core/src/generator.ts
ksaaskil: In Java or Scala world `CreateResponse` would, imho, return `Option<Response>`. In TypeScript we use `undefined` for the empty option. In my opinion it does not make sense to throw for an event that's very much expected to happen often (no match), but instead just return empty in that case and let the outer edge of the application handle that.

packages/unmock-core/src/generator.ts
idantene: I think it would be prettier and easier if we throw instead of return `undefined` :innocent:

ksaaskil:
But then one needs to add stuff like
```js
catch (err) {
  if (err.message === "SOMETHING_RANDOM_HERE") {
     return createDirectoriesAndNotifyUser()
  }
  throw err;
}
```
In my opinion returning undefined is much much cleaner for an expected event like that, showing that nothing unexpected went wrong in `unmock-js` but simply no match was found.

ksaaskil:
```js
const resOrUndefined = createResponse();
if (resOrUndefined === undefined) {
   handleNoMatch()
}
```
would be how the empty response is handled. To me that's not an error, that's something that is bound to happen.

packages/unmock-core/src/generator.ts
ksaaskil: Well I don't, this PR is not changing the existing logic so we can make another PR on it later on and let Mike decide.

packages/unmock-core/src/generator.ts
idantene: Well then the core issue is how often do we expect this to happen? I think it's not an expected outcome nor one that should happen often.
If you're using GitHub, you should have the GitHub service specification. If you're making a call to an endpoint that doesn't exist on GitHub - it's probably a typo.
If you're using an SDK and you have the service specification - I don't expect there to be any such mistakes (assuming the service specification is up to date).

The only time it would happen frequently IMO is when someone will want to build their service from the ground up?

ksaaskil: Yes, and we're still throwing an error for that event at the outer edge of our application. It's just that internally it's not an application error if no match is found so I don't see any reason to throw, it's user error and expected to happen.

idantene:
> But then one needs to add stuff like
> 
> ```js
> catch (err) {
>   if (err.message === "SOMETHING_RANDOM_HERE") {
>      return createDirectoriesAndNotifyUser()
>   }
>   throw err;
> }
> ```
> 
> In my opinion returning undefined is much much cleaner for an expected event like that, showing that nothing unexpected went wrong in `unmock-js` but simply no match was found.

I agree checking based on message is far from ideal. I imagine it more with a wrapper (if there's indeed a specific way a backend would want to treat a specific error, other than alerting the user):
```typescript
function getResponseOrHandleError() {
    try {
        return createResponse();
    } catch (e) {
        createDirectoriesAndNotifyUser();
        throw e;
    }
}

const response = getResponseOrHandleError();
```

ksaaskil:
> > But then one needs to add stuff like
> > ```js
> > catch (err) {
> >   if (err.message === "SOMETHING_RANDOM_HERE") {
> >      return createDirectoriesAndNotifyUser()
> >   }
> >   throw err;
> > }
> > ```
> > 
> > 
> > In my opinion returning undefined is much much cleaner for an expected event like that, showing that nothing unexpected went wrong in `unmock-js` but simply no match was found.
> 
> I agree checking based on message is far from ideal. I imagine it more with a wrapper (if there's indeed a specific way a backend would want to treat a specific error, other than alerting the user):
> 
> ```ts
> function getResponseOrHandleError() {
>     try {
>         return createResponse();
>     } catch (e) {
>         createDirectoriesAndNotifyUser();
>         throw e;
>     }
> }
> 
> const response = getResponseOrHandleError();
> ```

But that code still needs a check for if the error was "no match found" or something else, we don't want to start creating directories if unmock-js crashed.

packages/unmock-core/src/generator.ts
idantene: Well then maybe the best compromise would be to at least add type guards so that any backends that use `responseCreatorFactory` have to return a `ISerializedResponse`, so that any `undefined` is dealt with and not simply returned?

packages/unmock-core/src/generator.ts
idantene: I'd rather we merge so I can continue working and testing from `dev` without conflicts. We can change it, if needed, later on.

packages/unmock-core/src/generator.ts
ksaaskil: One cannot return `ISerializedResponse | undefined` from any backend, for Node.js one must construct valid `http.ServerResponse` from it and that's not possible to do without first checking if it's undefined. The compiler will barf with `Argument of type 'ISerializedResponse | undefined' is not assignable to parameter of type 'ISerializedResponse'.` if that check is removed.

PR #62, opened 2019-06-27
Track client requests to emit errors when things go wrong

ksaaskil:
- Simplified version of what [yesno](https://github.com/FormidableLabs/yesno) does

### DIFF
--- a/packages/unmock-node/src/__tests__/backend/index.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/index.test.ts
@@ -16,7 +16,7 @@ describe("Node.js interceptor", () => {
     afterAll(() => {
       nodeInterceptor.reset();
     });
-    test("gets successful response or petstore", async () => {
+    test("gets successful response for valid request", async () => {
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data).toBeDefined();
--- a/packages/unmock-node/src/__tests__/backend/index.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/index.test.ts
@@ -16,12 +16,12 @@ describe("Node.js interceptor", () => {
     afterAll(() => {
       nodeInterceptor.reset();
     });
-    test("gets successful response for petstore", async () => {
+    test("gets successful response or petstore", async () => {
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data).toBeDefined();
     });
-    test("gets emitted error for unknown url", async () => {
+    test("emits an error for unknown url", async () => {
       try {
         await axios("http://example.org");
       } catch (err) {
--- a/packages/unmock-node/src/__tests__/backend/index.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/index.test.ts
@@ -8,18 +8,27 @@ const servicesDirectory = path.join(__dirname, "..", "loaders", "resources");
 describe("Node.js interceptor", () => {
   describe("with petstore in place", () => {
     let nodeInterceptor: NodeBackend;
-    beforeEach(() => {
+    beforeAll(() => {
       nodeInterceptor = new NodeBackend({ servicesDirectory });
       const unmockOptions = new UnmockOptions();
       nodeInterceptor.initialize(unmockOptions);
     });
-    afterEach(() => {
+    afterAll(() => {
       nodeInterceptor.reset();
     });
-    test("gets successful response", async () => {
+    test("gets successful response for petstore", async () => {
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data).toBeDefined();
+    });
+    test("gets emitted error for unknown url", async () => {
+      try {
+        await axios("http://example.org");
+      } catch (err) {
+        expect(err.message).toBe("No matching template found");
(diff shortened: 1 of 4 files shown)


packages/unmock-node/src/__tests__/backend/index.test.ts
```diff
@@ -8,18 +8,27 @@ const servicesDirectory = path.join(__dirname, "..", "loaders", "resources");
...
   describe("with petstore in place", () => {
     let nodeInterceptor: NodeBackend;
-    beforeEach(() => {
+    beforeAll(() => {
       nodeInterceptor = new NodeBackend({ servicesDirectory });
       const unmockOptions = new UnmockOptions();
       nodeInterceptor.initialize(unmockOptions);
     });
-    afterEach(() => {
+    afterAll(() => {
```
ksaaskil: Cannot use `beforeEach` and `afterEach` here if tests run in parallel and share the same interceptor as the backend is stateful now. Not sure if this could break user tests as well in some cases.

packages/unmock-node/src/backend/index.ts
```diff
@@ -24,58 +34,148 @@ async function handleRequestAndResponse(
...
+
+  private extractTrackedClientRequest(
+    incomingMessage: IncomingMessage,
+  ): ClientRequest {
+    const { [UNMOCK_INTERNAL_HTTP_HEADER]: reqId } = incomingMessage.headers;
+    debugLog(
+      `Intercepted incoming request with ID ${reqId}, matching to existing IDs:`,
+      Object.keys(this.clientRequests),
+    );
+    if (typeof reqId !== "string") {
```
idantene: Is this the best way to check for non-empty request ID? I guess in theory `reqId` could be `""`?

packages/unmock-node/src/backend/index.ts
```diff
@@ -24,58 +34,148 @@ async function handleRequestAndResponse(
...
+        );
+        self.clientRequests[requestId] = { clientRequest: this };
+        this.setHeader(UNMOCK_INTERNAL_HTTP_HEADER, requestId);
+        return socket;
+      },
+    );
+
+    // Client-side socket connect, use to bypass connections
+    this.mitm.on("connect", (socket: IBypassableSocket, opts: RequestOptions) =>
+      this.mitmOnConnect.call(this, options, socket, opts),
```
idantene: Why the need to use `call` with `this` instead of `this.mitmOnConnect(options, socket, opts)`?

packages/unmock-node/src/backend/index.ts
```diff
@@ -24,58 +34,148 @@ async function handleRequestAndResponse(
...
+    this.origOnSocket = ClientRequest.prototype.onSocket;
+
+    /**
+     * When a socket is assigned to a client request, create an internal ID
+     * and add the ID in the request header. Thereby we can map
+     * the server-side incoming request to the corresponding
+     * client request and emit errors properly.
+     * Borrowed from https://github.com/FormidableLabs/yesno.
+     */
+    ClientRequest.prototype.onSocket = _.flowRight(
```
idantene: Just as a side note, as there are only two functions here, I'm sure we could get away without requiring `lodash` to pass value from one function to the other... :)
Is this how `yesno` does it as well?

packages/unmock-node/src/backend/index.ts
```diff
@@ -24,58 +34,148 @@ async function handleRequestAndResponse(
...
     }
     respondFromSerializedResponse(serializedResponse, res);
   } catch (err) {
-    // TODO Emit an error in the corresponding client request instead?
-    const errorResponse: ISerializedResponse = {
-      body: err.message,
-      statusCode: constants.STATUS_CODE_FOR_ERROR,
-    };
-    respondFromSerializedResponse(errorResponse, res);
+    clientRequest.emit("error", Error(`unmock error: ${err.message}`));
```
idantene: What do we suspect may throw here? `serializeRequest`?

packages/unmock-node/src/backend/index.ts
ksaaskil: It's just for type-safety, headers are typed as  `string`, `undefined` or `string[]` and handling 
`string[]` is overkill here I think.

packages/unmock-node/src/backend/index.ts
ksaaskil: Because in the latter case the function `this.mitmOnConnect(options, socket, opts)` would have `this` bound by the context, not necessarily being the class instance.

packages/unmock-node/src/backend/index.ts
ksaaskil: Yes that's how yesno does it, it's just a pretty way to do chaining.

packages/unmock-node/src/backend/index.ts
ksaaskil: Anything that can throw in the whole mock generation pipeline.

ksaaskil:
It might be better to keep the map of `request ID -> ClientRequest` in a singleton and to never reset that to avoid weird synchronization bugs, but I'll do that later on in another PR if it turns out to be a problem.

packages/unmock-node/src/backend/index.ts
ksaaskil: Actually it's true this does not need binding the context as it uses an arrow function and does not set the instance method as callback directly, I'll simplify this sometime.

PR #63, opened 2019-06-28
Service state validation

idantene:
- Verify content of given DSL against schema for specified service, in both nested and generic manner (`validator.ts`)
- Fetch relevant operations based on endpoint and method, including `DEFAULT_HTTP_METHOD` and `DEFAULT_ENDPOINT`, and their combinations (`service.ts`)
- Includes minimal parsing of DSL (at the moment only treats `$code`)
- Perks: Fixes typing error in parser, re-introduces `DEFAULT_HTTP_METHOD` as `ExtendedHTTPMethod` type, fixes logical error at `getAtLevel`
- Future PRs:
  - include options (skip runtime tests, match anything nested without specifying the entire response tree)
  - bootstrap to `Service`
- [x] Add tests
- [x] Simplify

### DIFF
--- a/packages/unmock-core/src/__tests__/service.util.test.ts
+++ b/packages/unmock-core/src/__tests__/service.util.test.ts
@@ -1,10 +1,13 @@
 import XRegExp from "xregexp";
-import { Parameter } from "../service/interfaces";
+import { Operation, Parameter, Response } from "../service/interfaces";
 import {
   buildPathRegexStringFromParameters,
+  findPathInObject,
   getAtLevel,
   getPathParametersFromPath,
   getPathParametersFromSchema,
+  getValidResponsesForOperationWithState,
+  verifyResponseWithState,
 } from "../service/util";
 
 describe("Tests getAtLevel", () => {
@@ -171,5 +174,37 @@ describe("Tests buildPathRegexStringFromParameters", () => {
         ["petId"],
       ),
     ).toThrow("following path parameters have not been described");
+  });
+});
+
+describe("Tests findPathInObject", () => {
+  const response: Response = {
+    content: {
+      "application/json": {
+        Pet: {
+          id: {
+            type: "integer",
+            format: "int64",
+          },
+          name: {
+            type: "string",
+          },
+          tag: { type: "string" },
+        },
+      },
+    },
+  };
+  it("with empty path", () => {
+    console.log(
+      verifyResponseWithState(response.content["application/json"], {
+        Pet: { id: "5", name: "bar" },
+      }),
+    );
+    console.log(
+      getValidResponsesForOperationWithState(
+        { responses: { 200: { ...response } } },
+        { Pet: { id: "5", name: "bar" }, $code: 200 },
+      ),
+    );
   });
 });
--- a/packages/unmock-core/src/__tests__/service.util.test.ts
+++ b/packages/unmock-core/src/__tests__/service.util.test.ts
@@ -1,13 +1,10 @@
 import XRegExp from "xregexp";
-import { Operation, Parameter, Response } from "../service/interfaces";
+import { Parameter } from "../service/interfaces";
 import {
   buildPathRegexStringFromParameters,
-  findPathInObject,
   getAtLevel,
   getPathParametersFromPath,
   getPathParametersFromSchema,
-  getValidResponsesForOperationWithState,
-  verifyResponseWithState,
 } from "../service/util";
 
 describe("Tests getAtLevel", () => {
@@ -174,37 +171,5 @@ describe("Tests buildPathRegexStringFromParameters", () => {
         ["petId"],
       ),
     ).toThrow("following path parameters have not been described");
-  });
-});
-
-describe("Tests findPathInObject", () => {
-  const response: Response = {
-    content: {
-      "application/json": {
-        Pet: {
-          id: {
-            type: "integer",
-            format: "int64",
-          },
-          name: {
-            type: "string",
-          },
-          tag: { type: "string" },
-        },
-      },
-    },
-  };
-  it("with empty path", () => {
-    console.log(
-      verifyResponseWithState(response.content["application/json"], {
-        Pet: { id: "5", name: "bar" },
-      }),
-    );
-    console.log(
-      getValidResponsesForOperationWithState(
-        { responses: { 200: { ...response } } },
-        { Pet: { id: "5", name: "bar" }, $code: 200 },
-      ),
-    );
   });
 });
--- a/packages/unmock-core/src/__tests__/service.util.test.ts
+++ b/packages/unmock-core/src/__tests__/service.util.test.ts
@@ -1,4 +1,5 @@
 import XRegExp from "xregexp";
+import { Parameter } from "../service/interfaces";
 import {
   buildPathRegexStringFromParameters,
   getAtLevel,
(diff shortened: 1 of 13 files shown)


packages/unmock-core/src/service/service.ts
```diff
@@ -55,58 +66,165 @@ export class Service implements IService {
...
+    debugLog(`Fetching operations for '${method} ${endpoint}'...`);
+    const { operations, error } = this.getOperations(method, endpoint);
+    if (error !== undefined) {
+      debugLog(`Couldn't find any matching operations: ${error}`);
+      return { success: false, error };
+    }
+    debugLog(`Found follow operations: ${operations}`);
+
+    // @ts-ignore TODO
+    operations.forEach((op: IOperationForStateUpdate) => {
```
idantene: Future PR: Bootstrapping it all together

packages/unmock-core/src/service/validator.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (content === undefined || content.schema === undefined) {
+      throw new Error(`No schema defined in ${JSON.stringify(content)}!`);
+    }
+    const schema = content.schema as Schema; // We assume '$ref' are already resolved at this stage
+    // For now, ignore $DSL notation and types
+    const responseModsForState = this.spreadStateFromService(
+      schema.properties,
+      state,
+    ) as Record<string, Schema>; // TODO is this the right type?
+    // TODO: Ignore DSL/convert elements
```
idantene: Future PR: Embed DSL logic, verify given value in `state` against fetched type in `responseMods`

packages/unmock-core/src/service/validator.ts
```diff
@@ -0,0 +1,187 @@
...
+   * to overwrite the default schema.
+   * The state items are matched by identifying all corresponding
+   * keys nested in the path leading to them.
+   *
+   * @param statePath A state being set in a service (or recursively-called, hence `any`)
+   * @param serviceSchema The top-level (or recursely-called, hence `any`) schema for the service
+   * @returns An object with string-keys leading to the location of matching state for each variable,
+   *          with value being either the Schema defined for that variable, or undefined.
+   */
+  public static spreadStateFromService(
```
idantene: Further simplified in #64

packages/unmock-core/src/service/validator.ts
```diff
@@ -0,0 +1,187 @@
...
+import {
+  isSchema,
+  MediaType,
+  Operation,
+  Response,
+  Schema,
+  UnmockServiceState,
+} from "./interfaces";
+
+export class StateRequestValidator {
```
idantene: This is far from ideal design, but it will be resolved in #63 - this is simply to get things going...

packages/unmock-core/src/__tests__/service.util.test.ts
```diff
@@ -109,41 +113,63 @@ describe("Tests getPathParametersFromSchema", () => {
...
+    { in: "path", name: "id" },
+    { in: "query", name: "foo" },
   ];
 
   it("Tests path", () => {
     // Empty schema - nothing to do
-    expect(buildPathRegexStringFromParameters("/pets", emptySchema, [])).toBe("/pets");
+    expect(buildPathRegexStringFromParameters("/pets", emptySchema, [])).toBe(
+      "/pets",
+    );
```
ksaaskil: Here it would be possible to follow the "one assert per test rule" by putting each of these expects in a unit test of their own, under a common `describe()`. But that's just a style thing.

packages/unmock-core/src/__tests__/validator.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  content: {
+    "application/json": {
+      schema,
+    },
+  },
+  description: "foo bar",
+};
+
+describe("Tests findStatePathInService", () => {
+  it("with empty path", () => {
```
ksaaskil: I think it would be quite helpful for future test readers if the name of the test told what the function is expected to do, something like `describe("finding state path in service")`,  `it("should produce empty state from empty path")`.

packages/unmock-core/src/service/constants.ts
```diff
@@ -1,5 +1,6 @@
 import XRegExp from "xregexp";
 
+export const DEFAULT_HTTP_METHOD = "any";
```
ksaaskil: I think this is a bit confusing: why is the "default HTTP method" of "service" equal to "any"? Isn't that part of the `dsl` and not inherently of service? Maybe we could consider isolating the state and DSL management to a directory of their own at some point...

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -17,21 +23,29 @@ const RESTMethodTypes = [
...
 export interface IServiceMapping {
   [serviceName: string]: IService;
 }
 
 export interface IStateInput {
-  method: HTTPMethod;
+  method: ExtendedHTTPMethod;
   endpoint: string;
-  newState: IUnmockServiceState;
+  newState: UnmockServiceState;
```
ksaaskil: Moving all the state stuff (in future) to a directory of their own would also shorten this file quite a bit.

packages/unmock-core/src/service/service.ts
```diff
@@ -55,58 +66,165 @@ export class Service implements IService {
...
 
-  public verifyRequest(method: HTTPMethod, endpoint: string) {
-    // Throws if method + endpoint are invalid for this service
+  public updateState({
+    method,
+    endpoint,
+    // @ts-ignore TODO
+    newState,
+  }: IStateInput): { success: boolean; error?: string } {
+    // Four possible cases:
```
ksaaskil: I think all such detailed documentation should rather be part of the JSDoc documentation for the method and not "inline" in the code?

packages/unmock-core/src/service/service.ts
```diff
@@ -55,58 +66,165 @@ export class Service implements IService {
     return this.matcher.matchToOperationObject(sreq);
   }
 
-  public verifyRequest(method: HTTPMethod, endpoint: string) {
-    // Throws if method + endpoint are invalid for this service
+  public updateState({
```
ksaaskil: Also here I would so much like if `service.updateState(input)` was just a simple call to `this.state.update(input)`, that would allow separating the concerns and produce more decoupled code.

packages/unmock-core/src/service/service.ts
idantene: They're remove in a future PR, these were internal notes :)

idantene:
Sure, I don't mind separating it into a `State` class. I don't think `Service` is becoming a god class though; note all the added private methods rarely depend on `Service`, mostly to avoid passing around the schema and/or name, and/or matcher.

I'll add that into #64 if that's okay with you. Same for separating the tests.

packages/unmock-core/src/service/validator.ts
ksaaskil: As TypeScript does not have static classes, could use `abstract` here to make it clear it's not supposed to be instantiated.

idantene: Removed from a class anyway (again, in #63) - it's not stateful so I don't see a point in it being a class...

ksaaskil: We're already using classes for non-stateful things like the matcher, I think it makes the architecture clear. But ofc it does not need to be a class.

idantene: True, I believe that's useful when there are multiple methods/functions being available. Here the end goal is just one function (with 2 more being exported for testing purposes), so I'd rather go with functions that are 'namespaced' by being in the same file.

PR #64, opened 2019-07-02
State bootstrap

idantene:
- [ ] Rebase onto `dev` once #63 is merged


packages/unmock-core/src/__tests__/validator.test.ts
```diff
@@ -40,29 +44,28 @@ const response: Response = {
...
       // Spreading from "test: { id : { ... " to also inlucde properties
-      test: {
-        properties: {
-          id: {
-            type: "integer",
-            format: "int64",
+      properties: {
+        test: {
+          properties: {
+            id: null, // Will be removed due to wrong type
```
idantene: Sets the value as needed. `null` is treated as a mismatch and is alerted when using higher-level functions.
`spreadStateFromService` is only exported for testing purposes here.

packages/unmock-core/src/service/state/utils.ts
```diff
@@ -0,0 +1,233 @@
...
+        Object.assign(filteredCodeToMedia, {
+          [mediaType]: codeSchema[mediaType],
+        }),
+      {},
+    );
+  }
+  return stateCodeToMedia;
+};
+
+/**
```
idantene: For brevity purposes, anything below this line is copied from PR #63 .

packages/unmock-core/src/service/state/validator.ts
```diff
@@ -0,0 +1,316 @@
...
+ *
+ * @param statePath A state being set in a service (or recursively-called, hence `any`)
+ * @param serviceSchema The top-level (or recursely-called, hence `any`) schema for the service
+ * @returns An object with string-keys leading to the location of matching state for each variable,
+ *          with value being either the Schema defined for that variable, or null.
+ * @throws  If a state path is defined in such a way, that the nested state isn't an object, string,
+ *          number or boolean (i.e. `{ path: { to: { state: undefined } } }` )
+ */
+// TODO: exported for testing purposes
+export const spreadStateFromService = (
```
mikesol: A few things:

- `servicePath` should have the type `UnmockServiceState`
- `UnmockServiceState` seems like it is pulling double duty of representing both a state transformation and a `$code`. If possible, I would separate these two things.
- Ideally, whatever winds up going to `servicePath` should be a function with the signature `Schema -> Schema`.
- Most of the content of `spreadStateFromService` can be one of several transformation functions that can be applied to a Schema.
- This would probably be a composable middleware pattern, ie `unmock.github(foo({id: 5}), bar('hello'), baz)` where `foo({id: 5})`, `bar('hello')` and `baz` are all functions with the signature `Schema -> Schema`.

packages/unmock-core/src/__tests__/state.test.ts
```diff
@@ -0,0 +1,209 @@
...
+      state.update({
+        stateInput: { method: "post", endpoint: "/test/5", newState: {} },
+        serviceName: "foo",
+        paths: fullSchema.paths,
+        schemaEndpoint: "/test/{test_id}",
+      }),
+    ).toContain("Can't find response for 'post /test/5'");
+  });
+
+  it("saves state from any endpoint and get method as expected", () => {
```
mikesol: Irrespective of the implementation, it would be good to write a test that modifies the top-level `id` without modifying `foo/id` and vice-versa.

idantene:
Perhaps `UnmockServiceState` is a bad naming decision, but the idea is that
it contains a state request, and that includes the DSL, hence it also
includes `$code`. These are parsed in the request and not saved as is in
the state itself.

`spreadStateFromService` is a single transformation IMO, not sure how most
of its content can be transformation functions... Either way, if we decide
down that route, I'd rather do that in a different PR.

For now I'd rather just cancel the nested search as we agreed on Slack if
that's fine by you?

mikesol:
Just to make sure we're on the same page, when you say in the PR response "cancel nested search", you mean that `{id: 5}` would only change a top-level property called `id` in content with a media type of `application/json`. So the default value assumes that the response body will be in a format `application/json` and will be an object instead of an array/string/etc?

And, for example, `{foo: {id: 5}}` would change the `foo.id` property if it exists?

Seems sensible as most APIs return JSON and most JSON is objects, but I just want to make sure that's what the idea is.

idantene:
Yes, without hard coding the "application/json" part (I guess the same
schema would apply to xml media type?)

packages/unmock-core/src/service/state/interfaces.ts
```diff
@@ -0,0 +1,29 @@
...
+  Schema,
+} from "../interfaces";
+
+export interface IStateUpdate {
+  stateInput: IStateInput;
+  serviceName: string;
+  paths: Paths;
+  /**
+   * Complements the endpoint given in IStateInput, by redirecting to the correct endpoint in Schema.
+   * This happens when an endpoint consists of variables, for example.
```
ksaaskil: I think an example would generally help to describe what the schema endpoint is. What's the difference between endpoint and a path in our code?

packages/unmock-core/src/service/service.ts
```diff
@@ -66,165 +41,40 @@ export class Service implements IService {
...
+    if (schemaEndpoint === undefined) {
+      throw new Error(`Can't find endpoint '${endpoint}' in '${this.name}'`);
     }
-    debugLog(`Found follow operations: ${operations}`);
 
-    // @ts-ignore TODO
-    operations.forEach((op: IOperationForStateUpdate) => {
-      // applyStateToOperation(op, newState);
-      // For each operation, verify the new state applies and save in `this.state`
+    const err = this.state.update({
```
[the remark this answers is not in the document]
ksaaskil: It looks a bit like C code to me 🙂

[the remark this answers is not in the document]
idantene: Just borrowing from your notion of not burying these too deep. Ofc `state.update` can throw itself. Will update to that.

packages/unmock-core/src/__tests__/state.test.ts
idantene: Done! (removed the nested search as agreed)

PR #71, opened 2019-07-04
Generator bootstrap

idantene:
- Binds the generator with the state and response.
- `responseCreatorFactory` returns the states object one can use to modify states for the `ServiceStore`.
- Adds `reset` methods to `ServiceStore`, `Service` and `State` (propagates down). Intended mostly for internal use when we'd like to reset the state after every test. Need to decide on a best method for this, as `states.reset()` and `states.github.reset()`, for example, mean that `reset` is a reserved word.
- Translated `default` response to `200`.
- [x] Add tests

### DIFF
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -1,4 +1,9 @@
-import { OpenAPIObject, PathItem, Schema } from "loas3/dist/src/generated/full";
+import {
+  OpenAPIObject,
+  Operation,
+  PathItem,
+  Schema,
+} from "loas3/dist/src/generated/full";
 import { ISerializedRequest } from "../interfaces";
 import { DEFAULT_STATE_HTTP_METHOD } from "./constants";
 
@@ -57,12 +62,19 @@ export interface IServiceInput {
   name: string;
 }
 
+//           (e.g. "application/json")   schema
+export type mediaTypeToSchema = Record<string, Schema>;
+//                              status   (e.g. "application/json")
+export type codeToMedia = Record<string, mediaTypeToSchema>;
+
 export interface IResponsesFromOperation {
-  // Maps between a status code, to a response type (e.g. "application/json") to a spread state
-  [statusCode: string]: Record<string, Record<string, Schema>>;
+  // Maps between a response method, to codeToMedia
+  [method: string]: codeToMedia;
 }
 
-export type MatcherResponse = any | undefined;
+export type MatcherResponse =
+  | { operation: Operation; state: codeToMedia | undefined }
+  | undefined;
 
 export interface IService {
   /**
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -100,6 +100,12 @@ export interface IService {
   updateState(input: IStateInput): void;
 
   /**
+   * Resets the entire state saved for this service.
+   * This provides easy access to wipe a state after every test.
+   */
+  resetState(): void;
+
+  /**
    * Match a given request to the service. Return an operation object for successful match,
    * undefined otherwise.
    * @param sreq Serialized request
--- a/packages/unmock-core/src/service/interfaces.ts
+++ b/packages/unmock-core/src/service/interfaces.ts
@@ -1,4 +1,4 @@
-import { OpenAPIObject, Schema } from "loas3/dist/src/generated/full";
+import { OpenAPIObject, PathItem, Schema } from "loas3/dist/src/generated/full";
 import { ISerializedRequest } from "../interfaces";
 import { DEFAULT_STATE_HTTP_METHOD } from "./constants";
 
@@ -33,6 +33,8 @@ const DEF_REST_METHOD = [DEFAULT_STATE_HTTP_METHOD] as const;
 type DEFAULT_HTTP_METHOD_AS_TYPE = typeof DEF_REST_METHOD[number];
 export type HTTPMethod = typeof RESTMethodTypes[number];
 export type ExtendedHTTPMethod = HTTPMethod | DEFAULT_HTTP_METHOD_AS_TYPE;
+
+export type OASMethodKey = keyof PathItem & HTTPMethod;
 
 export const isRESTMethod = (maybeMethod: string): maybeMethod is HTTPMethod =>
   RESTMethodTypes.toString().includes(maybeMethod.toLowerCase());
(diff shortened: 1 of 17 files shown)


packages/unmock-node/src/backend/index.ts
```diff
@@ -75,11 +75,16 @@ export default class NodeBackend implements IBackend {
...
   private origOnSocket?: (socket: net.Socket) => void;
   private mitm: any;
+  private stateStore: any;
 
   public constructor(config?: INodeBackendOptions) {
     this.config = { ...nodeBackendDefaultOptions, ...config };
   }
 
+  get states() {
+    return this.stateStore;
```
ksaaskil: I'd perhaps expect the initialize method to return this rather than have users access it on fly (with states being possibly undefined)?

packages/unmock-node/src/__tests__/backend/index.test.ts
```diff
@@ -8,19 +8,23 @@ const servicesDirectory = path.join(__dirname, "..", "loaders", "resources");
...
     afterAll(() => {
       nodeInterceptor.reset();
     });
+
     test("gets successful response for valid request", async () => {
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data).toBeDefined();
     });
+
```
ksaaskil: Could also add a few asserts for `response.data` to be of the correct format (if the generator now knows how to create proper data).

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -55,12 +62,19 @@ export interface IServiceInput {
   name: string;
 }
 
+//           (e.g. "application/json")   schema
+export type mediaTypeToSchema = Record<string, Schema>;
+//                              status   (e.g. "application/json")
```
ksaaskil: `mediaTypeToSchema` here is not `application/json` as the comment says but a map with possible key of `application/json`, I assume?

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -55,12 +65,12 @@ export class ServiceStore {
...
       throw new Error(
         `Can't find specification for service named '${
-          isExtendedRESTMethod(method) ? service : method
+          isExtendedRESTMethod(method) ? serviceName : method
         }'!`,
       );
     }
 
-    const error = this.serviceMapping[service].updateState({
+    const error = this.serviceMapping[serviceName].updateState({
```
ksaaskil: Is this still returning errors or throwing? 🤔

packages/unmock-node/src/backend/index.ts
```diff
@@ -75,11 +75,16 @@ export default class NodeBackend implements IBackend {
   } = {};
   private origOnSocket?: (socket: net.Socket) => void;
   private mitm: any;
+  private stateStore: any;
```
ksaaskil: How come this has no type?

packages/unmock-node/src/backend/index.ts
ksaaskil: Ok I see it's needed for modifying the state, personally I still it's a bit too tight coupling to give users access to something that's also directly used internally, I would have maybe expected the exposed thing to be something like `export const states = new DSL(backend.stateStore)` but it's not a big deal.

idantene: Sure! What would the wrapper do though?

packages/unmock-node/src/__tests__/backend/index.test.ts
idantene: :+1: tests are not finished just yet :)

packages/unmock-core/src/service/interfaces.ts
idantene: Yup, that's what I meant by the comment (`string` being a media type, e.g. `"application/json"`). Sorry for the lack of clarity there :)

idantene: Will update those comments to make it clearer.

packages/unmock-node/src/backend/index.ts
ksaaskil: Wrap calls to internal functions like `updateState` with DSL and proxy commands, but I guess that's what the state store is now doing? Forgot what that looks like...

packages/unmock-core/src/service/serviceStore.ts
idantene: Woops! Nice catch.

packages/unmock-node/src/backend/index.ts
idantene: `Proxy` returns type `any` AFAICS.

ksaaskil: Ok, did not realize that the state store is the proxy...

packages/unmock-node/src/backend/index.ts
ksaaskil: Checked out `stateStoreFactory`, that's pretty much doing what I was thinking to be the `new DSL` so nevermind!

packages/unmock-node/src/backend/index.ts
idantene: Would renaming this be beneficial (`stateStoreProxy`)?

packages/unmock-node/src/backend/index.ts
idantene: Returning it from call to `initialize()` is a good idea! Otherwise `states` is undefined.

packages/unmock-node/src/backend/index.ts
ksaaskil: Hmm not sure, maybe I just wasn't expecting a "store" to be the proxy handler for user-facing commands. But I don't have any suggestion for a better name.

packages/unmock-core/src/service/matcher.ts
```diff
@@ -130,32 +132,28 @@ export class OASMatcher {
...
     debugLog(`Searching for match for ${reqPath} for ${definedPaths}`);
 
     for (const pathItemKey of Object.keys(paths)) {
       const pathRegex = this.endpointToRegexMapping[pathItemKey];
-      if (pathRegex === undefined) {
-        continue;
-      }
-      if (pathRegex.test(reqPath)) {
-        return pathItemKey;
+      if (XRegExp.test(reqPath, pathRegex)) {
```
idantene: Figured this can't be undefined as we build `endpointToRegexMapping` for every `Object.keys(paths)`

PR #72, opened 2019-07-05
Refactor the tracking of client requests

ksaaskil:
- Refactor client request tracking to a separate `ClientRequestTracker` class
- Keep tracking logic in a static object to avoid problems with multiple backends existing simultaneously (maybe the backend should be forced to be a singleton in any case) and being initialized and reset independently
- Delete outgoing requests from the map once the request is intercepted

### DIFF
--- a/packages/unmock-node/src/backend/client-request-tracker.ts
+++ b/packages/unmock-node/src/backend/client-request-tracker.ts
@@ -76,7 +76,7 @@ export default abstract class ClientRequestTracker {
    * Deletes the corresponding instance from the map of tracked requests.
    * @param incomingMessage Incoming message ("server" side)
    */
-  public static extractTrackedClientRequest(
+  public static popTrackedClientRequest(
     incomingMessage: IncomingMessage,
   ): ClientRequest {
     const { [UNMOCK_INTERNAL_HTTP_HEADER]: reqId } = incomingMessage.headers;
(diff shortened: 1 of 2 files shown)


packages/unmock-node/src/backend/client-request-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const UNMOCK_INTERNAL_HTTP_HEADER = "x-unmock-req-id";
+
+/**
+ * "Static" class for tracking client requests.
+ */
+export default abstract class ClientRequestTracker {
+  public static readonly outgoingRequests: {
+    [requestId: string]: {
+      clientRequest: ClientRequest;
```
idantene: BTW is this in an object (as opposed to `[requestId: string]: ClientRequest`) in case we'd like to store more items here in the future?

packages/unmock-node/src/backend/client-request-tracker.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Stop tracking client requests, restore `ClientRequest.prototype.onSocket`.
+   */
+  public static stop() {
+    if (!ClientRequestTracker.active) {
+      return;
+    }
+
+    if (!ClientRequestTracker.origOnSocket) {
```
idantene: Isn't this checked with `ClientRequestTracker.active` (checks if `ClientRequestTracker.origOnSocker !== undefined`)?

ksaaskil: Yes but TypeScript does not realize that's a type-check, maybe the return type could be some kind of an `is` check...

packages/unmock-node/src/backend/client-request-tracker.ts
ksaaskil: Yeah it's a remnant of yesno as they keep track of more objects, could remove one level of nesting there.

PR #73, opened 2019-07-08
State validation as function

idantene:
Changes the state passed to a `IStateInputGenerator` object with 1 function and 2 properties:
1. `gen(schema: Schema)` consumes a `Schema` and generates the expanded matching state based on the schema, filtering out the top-level DSL.
2. `top` contains only the top-level DSL from the given state.
3. `isEmpty` boolean denoting whether or not the state is empty, to avoid generation when unneeded.

The methods wrapping the states are known as `providers`.
Also removes `getUpdatedStateFromContent` and moves the content to `getStateFromMedia` (reducing exports and number of calls), and migrates matching tests.

- [x] Define the default provider to be the current state-schema "unraveler", so that users do not have to specify that if they don't want to.
- [x] Migrate tests
- [x] Add tests for providers

### DIFF
--- a/packages/unmock-core/src/service/state/validator.ts
+++ b/packages/unmock-core/src/service/state/validator.ts
@@ -146,9 +146,10 @@ const getStateFromMedia = (
     const content = contentRecord[contentType];
     if (content === undefined || content.schema === undefined) {
       errors.push({
-        msg: `No schema defined in ${JSON.stringify(content)}!`,
+        msg: `No schema defined in '${JSON.stringify(content)}'!`,
         nestedLevel: -1,
       });
+      continue;
     }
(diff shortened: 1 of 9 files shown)


packages/unmock-core/src/service/state/validator.ts
```diff
@@ -144,50 +144,26 @@ const getStateFromMedia = (
...
   return { responses: relevantResponses, errors };
 };
 
-/**
- * Given a media type, return mapping between state, service and response.
- * @param content
- * @param state
- */
-// TODO: exported for testing purposes
-export const getUpdatedStateFromContent = (
```
idantene: Embedded in `getStateFromMedia`, tests migrated as needed.

PR #76, opened 2019-07-10
[DSL] Translating $size (non top level DSL)

idantene:
- Translate `$size` argument to `minItems` and `maxItems`. This also reduces the responsibility of the generator since everything is OAS compliant.
- Handles inconsistency (i.e. `$size` given to non-array types, or invalid value for `$size` is given) based on `DSL.STRICT_MODE`. When `true`, throws an error. When `false`, treat "as is" (i.e. non-DSL argument).
- Splits subpackage to different responsibilities (actors and translators).
- [x] Add tests
- [x] Rebase after merging #74

### DIFF
--- a/packages/unmock-core/src/service/dsl/dsl.ts
+++ b/packages/unmock-core/src/service/dsl/dsl.ts
@@ -1,93 +1,17 @@
 import debug from "debug";
 import { cloneDeep } from "lodash";
-import {
-  codeToMedia,
-  isReference,
-  mediaTypeToSchema,
-  Schema,
-} from "../interfaces";
+import { codeToMedia, Schema } from "../interfaces";
+import { actOn$times } from "./actors";
+import { SCHEMA_TIMES } from "./constants";
 import { ITopLevelDSL } from "./interfaces";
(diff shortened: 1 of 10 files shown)


packages/unmock-core/src/service/dsl/dsl.ts
```diff
@@ -1,42 +1,48 @@
...
+ * Many of these static functions modify their inputs to abstract away
+ * the DSL from the final output (after having it translated/acted upon).
+ */
+export abstract class DSL {
+  /**
+   * If true, mismatching arguments will throw
+   *    (e.g. `$size: a`, `$size: 0.1` (rounds to 0), `$size: N` with non-array type)
+   * If false, these arguments are considered "as if" when they don't match the schema.
+   */
+  public static STRICT_MODE = true;
```
idantene: Not sure if we want to expose this to end users. I put it here for the time being so it's easy to switch between the modes internally and decide what fits best for the flow.
We could expose it a-la `unmock.strict_dsl()` or `unmock.lax_dsl()`?

PR #81, opened 2019-07-12
Add text middleware

idantene:
- Adds (badly named) `textMW` to set string-based states. Accepts a string argument and optional top-level DSL to ask for e.g. status code response.
  - This, in theory, lays the basis for possible future middlewares such as binary, file, etc (in _Open API schema-speak_, these are all schemas of `type: string` and different `format`). As a result, should these middlewares be inheritable classes?
- Cleans up some old code in validator and generator, that would otherwise crash on text results (both purposefully originally and unintentionally).
- Adds tests

### DIFF
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -101,6 +101,20 @@ const getStateForOperation = (
   };
 };
 
+/**
+ * Provides a work-around with for functions that may fail with a default value.
+ * Attemps to return `f(value)`. If an error is thrown, returns `value`.
+ * @param value
+ * @param f
+ */
+const tryCatch = (value: any, f: (value: any) => any) => {
+  try {
+    return f(value);
+  } catch {
+    return value;
+  }
+};
+
 const chooseResponseFromOperation = (
   operation: Operation,
 ): { $code: string; template: Schema } => {
@@ -152,15 +166,10 @@ const generateMockFromTemplate = (
   // First iteration simply parses these and returns the updated schema
   const resolvedTemplate = jsf.generate(template);
   jsf.reset();
-  // After one-pass resolving we might have new parameters to resolve.
-  let body: string;
-  try {
-    body = JSON.stringify(jsf.generate(resolvedTemplate));
-  } catch {
-    body = JSON.stringify(resolvedTemplate);
-  }
 
-  // 5. Generate as needed
+  // After one-pass resolving we might have new parameters to resolve.
+  const body = JSON.stringify(tryCatch(resolvedTemplate, jsf.generate));
+
   return {
     body,
     // TODO: headers
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -152,10 +152,17 @@ const generateMockFromTemplate = (
   // First iteration simply parses these and returns the updated schema
   const resolvedTemplate = jsf.generate(template);
   jsf.reset();
+  // After one-pass resolving we might have new parameters to resolve.
+  let body: string;
+  try {
+    body = JSON.stringify(jsf.generate(resolvedTemplate));
+  } catch {
+    body = JSON.stringify(resolvedTemplate);
+  }
 
   // 5. Generate as needed
   return {
-    body: JSON.stringify(jsf.generate(resolvedTemplate)),
+    body,
     // TODO: headers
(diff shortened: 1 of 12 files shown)


packages/unmock-core/src/service/state/middleware.ts
```diff
@@ -17,6 +17,19 @@ export default (state?: UnmockServiceState): IStateInputGenerator => ({
     spreadStateFromService(schema, filterTopLevelDSL(state || {})),
 });
 
+export const textMW = (state?: string): IStateInputGenerator => ({
```
idantene: Far from ideal naming, consider this a temporary name for now...

packages/unmock-core/src/generator.ts
```diff
@@ -152,10 +152,17 @@ const generateMockFromTemplate = (
   // First iteration simply parses these and returns the updated schema
   const resolvedTemplate = jsf.generate(template);
   jsf.reset();
+  // After one-pass resolving we might have new parameters to resolve.
+  let body: string;
+  try {
+    body = JSON.stringify(jsf.generate(resolvedTemplate));
+  } catch {
```
idantene: This happens on textual responses (when trying to `jsf.generate` from non-object).

packages/unmock-core/src/service/state/middleware.ts
```diff
@@ -62,7 +75,9 @@ export const spreadStateFromService = (
         // TODO do we want to throw for invalid types?
         const spread = {
           [key]:
-            isSchema(scm) && ajv.validate(scm, stateValue) ? stateValue : null,
+            isSchema(scm) && ajv.validate(scm, stateValue)
+              ? { ...scm, const: stateValue }
```
idantene: `const` is not yet represented in `loas` types but it's officially part of JSON schema. This feels like the "correct" way to do this, rather than override the schema for a given key.

packages/unmock-node/src/__tests__/backend/index.test.ts
```diff
@@ -25,11 +25,12 @@ describe("Node.js interceptor", () => {
       const data = response.data;
       // As no specific code was set, we expect either valid response
       // or an error response (based on service specification)
+
       if (data.message !== undefined) {
         // error message chosen at random
         expect(typeof data.code === "number").toBeTruthy();
         expect(typeof data.message === "string").toBeTruthy();
-      } else {
+      } else if (typeof data !== "string") {
```
idantene: `data` could now be `text/plain` too (after modifying `spec.yaml`)

packages/unmock-core/src/generator.ts
```diff
@@ -152,10 +152,17 @@ const generateMockFromTemplate = (
   // First iteration simply parses these and returns the updated schema
   const resolvedTemplate = jsf.generate(template);
   jsf.reset();
+  // After one-pass resolving we might have new parameters to resolve.
```
mikesol: For this type of try/catch pattern, it would be easier to read if it used `Maybe` from `fp-ts` or a similar type of structure so that `body` could be a `const`.

packages/unmock-core/src/service/state/middleware.ts
```diff
@@ -17,6 +17,22 @@ export default (state?: UnmockServiceState): IStateInputGenerator => ({
     spreadStateFromService(schema, filterTopLevelDSL(state || {})),
 });
 
+export const textMW = (
```
mikesol: If this is a general middleware pattern (meaning if the function signature is supposed to be the same for all middleware), could we use that here?

idantene: The return type needs to be the same (`IStateInputGenerator`), but the signature itself (wrt to input) can be different as needed.

packages/unmock-core/src/generator.ts
idantene: Sure! I'm not sure it's more readable though, I ended up with:
```typescript
const body = JSON.stringify(
  getOrElse(() => resolvedTemplate)(
    tryCatch(() => jsf.generate(resolvedTemplate)),
  ),
);```

idantene: Decided to instead implement an abstraction a-la:
```typescript
const tryCatch = (v: any, f: (v: any) => any) => {
  try {
    return f(v);
  } catch {
    return v;
};```

PR #82, opened 2019-07-15
General spec name

idantene:
- Remove the strict requirement for spec files to be `spec`, `index`, etc, and allow JSON files as well.
- Verify if they're valid OAS by `isOpenAPIObject` from `loas3`.
Not sure at this point if it's worthwhile to try parsing with `loas3` or checking with `isOpenAPIObject`... :thinking:


idantene:
Will open a cleaner PR following the dev update

PR #95, opened 2019-07-22
Deref only when needed

idantene:
- Instead of dereferencing entire schema and service specs, only deref when needed - in other words, only when setting a state and verifying it's correct.
- Only dereferences basic references - using local relative/absolute URIs, or linking to other local files. Does not resolve `$id` or remote files (will be left for another PR).
- This entails, at the moment (open to suggestions) - passing along a curried function to many methods (propagating down/up as needed).
- Changes nothing from the user perspective. 
- [ ] Add more tests for dereferencing.

### DIFF
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -16,14 +16,16 @@ import {
   codeToMedia,
   Header,
   IService,
-  isReference,
   MatcherResponse,
   Operation,
+  Response,
   Responses,
   Schema,
+  ISchemaForDeref,
 } from "./service/interfaces";
 import { ServiceParser } from "./service/parser";
 import { ServiceStore } from "./service/serviceStore";
+import { derefIfNeeded } from "./service/util";
 
 type Headers = Record<string, Header>;
 
@@ -70,6 +72,7 @@ const setupJSFUnmockProperties = () => {
 const getStateForOperation = (
   operation: Operation,
   state: codeToMedia | undefined,
+  derefSchema: ISchemaForDeref,
 ):
   | {
       $code: string;
@@ -93,14 +96,11 @@ const getStateForOperation = (
   }
 
   const operationResponse = responses[statusCode as keyof Responses];
-  if (
-    operationResponse === undefined ||
-    isReference(operationResponse) ||
-    operationResponse.content === undefined
-  ) {
+  if (operationResponse === undefined) {
     return undefined;
   }
-  const operationContent = operationResponse.content;
+  const resolvedResponse = derefIfNeeded(operationResponse, derefSchema);
+  const operationContent = resolvedResponse.content;
 
   const mediaTypes = Object.keys(state[statusCode]).filter((type: string) =>
     Object.keys(operationContent).includes(type),
@@ -116,7 +116,7 @@ const getStateForOperation = (
   return {
     $code: statusCode,
     template: defaultsDeep(requestedState, matchedOperation),
-    headers: operationResponse.headers as Headers,
+    headers: resolvedResponse.headers,
   };
 };
 
@@ -136,6 +136,7 @@ const tryCatch = (value: any, f: (value: any) => any) => {
 
 const chooseResponseFromOperation = (
   operation: Operation,
+  derefSchema: ISchemaForDeref,
 ): {
   $code: string;
   template: Schema;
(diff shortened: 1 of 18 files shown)


packages/unmock-core/src/service/interfaces.ts
```diff
@@ -106,6 +109,16 @@ export interface IService {
...
 
+  /**
+   * Holds the absolute path where the service specification resides.
+   */
+  readonly absPath: string;
+
+  /**
+   * Used to dereference items in the schema when needed
+   */
+  readonly dereferencer: Dereferencer;
```
ksaaskil: Do we expect different kinds of dereferencers instead of a single function?

ksaaskil: Why does this need to be exposed by the service? 🤔

packages/unmock-core/src/service/util.ts
```diff
@@ -1,6 +1,51 @@
...
+  const isLocal = refValue.startsWith("#");
+  const afterPound = refValue
+    .split("#")
+    .slice(1)
+    .join("#");
+
+  const schema = isLocal
+    ? derefSchema.schema
+    : fs.readFileSync(
+        fspath.join(derefSchema.absPath, refValue.split("#")[0]),
```
ksaaskil: Should one perhaps check if the file referenced by the reference even exists? Could it be something else than a file as well?

packages/unmock-core/src/generator.ts
```diff
@@ -148,28 +148,31 @@ const chooseResponseFromOperation = (
   }
 
   const response = responses[chosenCode as keyof Responses];
-  if (
-    response === undefined ||
-    isReference(response) ||
-    response.content === undefined
-  ) {
+  if (response === undefined) {
+    throw new Error("Not sure what went wrong");
```
ksaaskil: Maybe a real error message like `Got undefined response` would be easier to debug?

packages/unmock-core/src/service/interfaces.ts
idantene: It's used with the generator - hence the public attribute.
Not expecting different kinds of dereferencers for now. It mostly makes it easier to mock. We could expand on this though (see my other replies).

packages/unmock-core/src/service/util.ts
idantene: It can most definitely be something other than a file. In theory, it could also be e.g. a web URL, a link to `$id`, etc. There are many options and I'm not sure if we want to implement all of them for now - or at least in this PR. They could also be a subject of different "dereferencers". I can see that being set in e.g. `UnmockOptions`.

If the file does not exist - it will throw and I think it's a quite verbose error when a file is not found.

packages/unmock-core/src/generator.ts
idantene: :+1: I can change to that. Many of those are mostly for type-checking, but I don't see why e.g. `const response = responses[chosenCode as keyof Responses]` will ever be undefined when `chosenCode` is defined and chosen from the keys of `responses`, etc.

packages/unmock-core/src/service/util.ts
ksaaskil: Ok. I don't really understand how different dereferencers would work, afaics there's only one way to resolve a reference and that's by following the reference. Also don't see why it needs to be mocked 🤔 I can see issues where the dereferencing follows external URLs but not sure if that would be easier to mock by patching instead of passing the dereferencer around...

idantene: In theory, one could pass e.g. `offlineMode` as parameters and then fail when trying to resolve URLs fetches. But like I said - I don't _really_ expect different kinds of dereferencers.

It only needs to be mocked when e.g. testing a `State` that was not initialized by a `Service` (then the `dereferencer` is not initialized, but `state.update` still expects it). Same when checking other methods that use a `dereferencer`.

ksaaskil: Cool. I kinda wouldn't have expected it to be a service's job to keep track of how references should be resolved for other users of the schema but maybe exposing a decorated schema that is aware of how to resolve those. But I don't understand the full architecture well enough to know which one is best, does not really matter.

packages/unmock-core/src/service/util.ts
```diff
@@ -1,6 +1,51 @@
...
 import XRegExp from "xregexp";
 import { OAS_PATH_PARAM_REGEXP, OAS_PATH_PARAMS_KW } from "./constants";
-import { Parameter, Paths } from "./interfaces";
+import { OpenAPIObject, Parameter, Paths } from "./interfaces";
+
+/**
+ * Simple dereferencing JSON schema $ref with JSON pointers or URIs.
+ * See https://json-schema.org/understanding-json-schema/structuring.html for more.
+ */
+export function derefIfNeeded(
```
ksaaskil: To me it might be clearer if this was curried, with type `derefIfNeeded = ({ schema, absPath }) => (reference: any) => any` but it's a matter of taste.

idantene: True! Forgot to update that, thanks!

PR #97, opened 2019-07-23
Renames middleware to transformers and textMW to textResponse

idantene:
Would probably make things more readable easier to understand.

### DIFF
--- a/packages/unmock-core/src/__tests__/state.test.ts
+++ b/packages/unmock-core/src/__tests__/state.test.ts
@@ -3,8 +3,8 @@ import {
   HTTPMethod,
   IStateInputGenerator,
 } from "../service/interfaces";
-import defMiddleware, { textMW } from "../service/state/middleware";
 import { State } from "../service/state/state";
+import defMiddleware, { textResponse } from "../service/state/transformers";
 
 const fullSchema = {
   openapi: "",
@@ -272,7 +272,7 @@ describe("Test state management", () => {
 
   it("Handles textual state correctly", () => {
     const state = new State();
-    updateState(state, "any", "**", textMW("foobar"), "**");
+    updateState(state, "any", "**", textResponse("foobar"), "**");
     const stateResult = getState(state, "get", "/");
     expect(stateResult).toEqual({
       200: { "text/plain": { const: "foobar", type: "string" } },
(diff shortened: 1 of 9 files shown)


packages/unmock-core/src/__tests__/state.test.ts
```diff
@@ -272,7 +272,7 @@ describe("Test state management", () => {
 
   it("Handles textual state correctly", () => {
     const state = new State();
-    updateState(state, "any", "**", textMW("foobar"), "**");
+    updateState(state, "any", "**", textResponse("foobar"), "**");
```
ksaaskil: Not related to this PR, but this looks quite hard to read and error-prone with so many arguments...

idantene: I understand, but it's a shortcut I made for internal tests (`updateState` is defined locally in this file as a shorthand for an otherwise vertically-breaking code).

idantene:
> Hmm, I'm not really sure users need or want to access an object called "transformers". What does it transform and why? Maybe I would rather expect to see something like `import { textResponse } from "unmock-node/dsl";` or something...

I wanted to separate the "transformers" from unmock page alone. It can be part of DSL, of course. Do you have a better naming suggestion for it, other than transformers?

packages/unmock-core/src/__tests__/state.test.ts
ksaaskil: Sure, don't see though why internal helpers should not have the same standards as exposed functions :)

idantene: Ah, just because in this case it loses it's function. I wanted to stop vertical line break, and making it e.g. a single object with parameters will again cause multiple line breaks.
If it bothers you I can revert back and eliminate that though.

ksaaskil: It doesn't bother me, just a comment.

ksaaskil:
It can be a transformer in the internal code, I would just avoid exposing too complex(-sounding) APIs.

idantene:
Fair, I have #98 (soon to be a PR) with revising import/export statements etc (treating one of Mike's learnings from the Odessa workshops about importing unmock). I'll update it there.

PR #98, opened 2019-07-23
Revise import notation

idantene:
- `unmock-node` to only export `unmock` object (inherits from abstract class `CorePackage` - used to make sure we share interface in other modules if/when needed). Changes import statement to `import unmock from "unmock-node";` (or anything else, using default export) 
- Exports `dsl` as a named attribute. If we'd like to have `import { textResponse } from "unmock-node/dsl";` we need to both publish only the `dist` dir of every package, and create that directory in every package. For now it's exported as-is.
- Renames the default "transformer" as `objResponse`. This change affects some test files.


package.json
```diff
@@ -5,7 +5,7 @@
     "compile": "tsc --build tsconfig.build.json",
     "compile:clean": "tsc --build tsconfig.build.json --clean",
     "watch": "tsc --build tsconfig.build.json --watch",
-    "release": "npm run clean && npm install && lerna publish --exact --force-publish=unmock-core --include-merged-tags",
+    "release": "npm run clean && npm install && lerna publish from-package --exact --force-publish=unmock-core --include-merged-tags",
```
ksaaskil: Not related to this PR, but why does it have `--force-publish=unmock-core`?

idantene: No idea. Maybe @mikesol could shed some light on that :)
Maybe it was a safeguard or something.

idantene:
@ksaaskil This also means that in the documentation one can now `import unmock from "unmock-node"` and not use the asterisk notation.

ksaaskil:
Looks good 👍

PR #99, opened 2019-07-23
Adds typing for state store

idantene:
Adds hinting and typing for unmock-related state store management. This includes suggestions on the fluent API and inputs when setting a state.

The main caveat at the moment is that the REST methods are manually typed (including their `JSDoc`) so that they get a method type and not a property type.

I've checked various edge cases but please pull and try it out, maybe break it with various calls via e.g. `const myTestStore = stateStoreFactory(new ServiceStore([]));`.

### DIFF
--- a/packages/unmock-core/src/service/index.ts
+++ b/packages/unmock-core/src/service/index.ts
@@ -3,8 +3,8 @@ import {
   ExtendedHTTPMethod,
   HTTPMethod,
   isRESTMethod,
-  UnmockServiceState,
   IStateInputGenerator,
+  UnmockServiceState,
 } from "./interfaces";
 import { ServiceStore } from "./serviceStore";
 export { ServiceParser } from "./parser";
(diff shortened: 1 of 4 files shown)


MeeshkanCain:
image

As an example :arrow_up: 

(Some pet peeve about showing `get, post, put` etc as properties...)

packages/unmock-core/src/service/index.ts
```diff
@@ -3,6 +3,7 @@ import {
   ExtendedHTTPMethod,
   HTTPMethod,
   isRESTMethod,
+  StateStoreType,
```
ksaaskil: I still don't like the name `Store` for the exposed services, to me "store" is something else. But I don't really have a better alternative, maybe `StateFacade` or something but not sure that really matters at all.

idantene: Sure, naming is often my weak spot.

PR #100, opened 2019-07-24
change tslint to prettier

boubamaraka:


mikesol:
Almost there! It is still breaking the build because `lint-ts` is being called somewhere. You'd need to replace that with `lint`.

PR #102, opened 2019-07-24
Fix small oversights

idantene:
- Forgot to update states returned from e.g. `responseCreatorFactory` to use the new types from #99 
- Updates `matcher` to replace server path prefix fully and not by partial match. Error rose when trying to set state for path `/api.test` when the server prefix is `/api`. Instead of replacing `/api` with the empty string, replaces `/api/` with `/` instead.

### DIFF
--- a/packages/unmock-core/src/__tests__/serviceStore.test.ts
+++ b/packages/unmock-core/src/__tests__/serviceStore.test.ts
@@ -63,58 +63,60 @@ describe("Fluent API and Service instantiation tests", () => {
 
   it("Store with existing path does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
-    store.petstore(); // Should pass
+    store.petstore({}); // Should pass
   });
 
   it("Store with REST method call does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
-    store.petstore.get(); // Should pass
+    store.petstore.get({}); // Should pass
   });
 
   it("Chaining multiple states without REST methods does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
     store
-      .petstore()
-      .petstore()
-      .petstore();
+      .petstore({})
+      .petstore({})
+      .petstore({});
   });
 
   it("Chaining multiple states with REST methods does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
     store.petstore
-      .get()
-      .petstore.get()
-      .petstore();
+      .get({})
+      .petstore.get({})
+      .petstore({});
   });
 
   it("Chaining multiple methods for a service does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
     store.petstore
-      .get()
-      .get()
-      .petstore();
+      .get({})
+      .get({})
+      .petstore({});
   });
   it("Non HTTP methods are recognized as services and throws", () => {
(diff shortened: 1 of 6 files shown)


packages/unmock-core/src/__tests__/serviceStore.test.ts
```diff
@@ -63,58 +63,60 @@ describe("Fluent API and Service instantiation tests", () => {
 
   it("Store with existing path does not throw", () => {
     const store = stateFacadeFactory(PetStoreWithPseudoResponses);
-    store.petstore(); // Should pass
+    store.petstore({}); // Should pass
```
ksaaskil: Hmm, why didn't this throw an error before if `store.petstore()` does not make sense? 🤔

idantene: I updated it just to fix the type hinting, behind the scenes an empty object is treated the same as undefined object (no-op). Not sure if it should throw.

ksaaskil: Hmm, I don't understand what there was to "fix" if users can anyway do either 🤔

MeeshkanCain: New types mean vscode underlines such mistakes (even though both are acceptable) - 
image

Alternatively ofc I could include 0 arguments as valid, if you feel that's more correct.

ksaaskil: No, just wondering why `tsc` compiled if it's wrong, maybe we don't even do type checking for the test files.

idantene: It'll now attempt to compile tests as well.

ksaaskil: Cool, I'll probably have to test that out at some point as I don't understand how adding `tsconfig.json` to the test folders adds the type checking, unless that's some lerna magic...

idantene: Ah, some lerna magic. Running `tsc --build tsconfig.test.json` runs `tsc` in all the given `paths` in that file and looks for `tsconfig.json` in each. If it doesn't find it - it crashes.
So of the two options - commenting packages out or adding a boilerplate, I chose the latter :)

ksaaskil: Cool, do we run that anywhere in CI/CD?

idantene: Yup, we use `npm ci` in CI, which implicitly calls `postinstall`, which calls `compile` :innocent:

ksaaskil: Interesting, if compiling test and source code happen in the same step I sure hope lerna hows to keep the build clean!

idantene: That's taken care of by `tsc --build tsconfig.test.json --clean`, which removes the compiled test files.

ksaaskil: Ok, I'm still so suspicious that I'll probably have to test adding a type check error in the test files and see what happens myself but this PR is of course very good to go!

idantene: Okay I just noticed I forgot the update to `package.json`, so that clarifies all this confusion hopefully. :man_facepalming: My bad.

PR #103, opened 2019-07-24
Accept text response natively

idantene:
- Parses string states as `textResponse` by default.
- Leaves `objResponse` and `textResponse` as exports under hoisted `transformers` as `dsl`.
- Updates tests to reflect import changes.

### DIFF
--- a/packages/unmock-node/src/__tests__/backend/states.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/states.test.ts
@@ -1,6 +1,7 @@
 import axios from "axios";
 import path from "path";
-import { CorePackage, dsl, UnmockOptions } from "unmock-core";
+import { CorePackage, UnmockOptions } from "unmock-core";
+import { dsl } from "../..";
 import NodeBackend from "../../backend";
 
 class StateTestPackage extends CorePackage {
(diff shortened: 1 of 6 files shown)


packages/unmock-node/src/__tests__/backend/states.test.ts
```diff
@@ -77,13 +78,20 @@ describe("Node.js interceptor", () => {
...
-      states.petstore(dsl.textResponse("foo"));
+    test("gets correct state when setting textual response", async () => {
+      states.petstore("foo");
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data).toBe("foo");
     });
 
+    test("gets correct state when setting textual response with path", async () => {
+      states.petstore("/pets", "bar");
```
ksaaskil: How do you set the status code here then? Using `textResponse`? I'll say again that to me `states.petstore("foo");` is not very readable  🧙‍♂️

idantene: Yup, `textResponse` if you have an additional, "top level DSL" instruction you need to use (`$times`, `$code`, more-to-come).

I assumed it's clearer (maybe not with `"foo"` though, admittedly), as -
> ... states.github("user logged in") ..., to me `states.github` says that you’re modifying the state for `github` and setting the state to `user logged in`

But so are you suggesting that we should be able to parse it only if 2 arguments are given? Or are you against the default `**` path?

ksaaskil: No suggestion specifically, that's just a drawback of not using longer state input objects like `{ response: { body: "foo" } }` but there's really nothing we can do about it, only deciding what syntax we want to encourage users to use.

idantene: We can always change the syntax later or decide on DSL instructions, or "transformers", or any combination of them (or all of them!).

I think these needs will come from the different PRs we're making to other repos, seeing what the use-cases are, how using `unmock` changes testing patterns, etc.

Personally, I prefer shorter syntax to more verbose one, but I do understand that's strictly my own preference (which might be a bad habit).

ksaaskil: Short is good, personally I much prefer the readable one though as that's easier to maintain and easier for developers possibly unfamiliar with the project to understand. It would be a different story if I was defining behaviour multiple times a day for unmock service. But yeah, working on the new PRs is good.

PR #104, opened 2019-07-24
Release/v0.1.4

idantene:
To test the flow of pushing a new version, new CircleCI configuration, etc.
- Bumping version with lerna pushses a tag by default. Instead, run `npm run version` (or `npx lerna version --no-git-tag-version`).
- [x] Bump version -> adds tag by default -> remove tag and update package.json
- [x] Update changelog
- [x] Add script to enforce updates to changelog and version bump

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,17 +5,55 @@
 version: 2.1
 orbs:
   codecov: codecov/codecov@1.0.2
+
+defaults: &defaults
+  working_directory: ~/repo
+  docker:
+    - image: circleci/node:10.15.1
+
 jobs:
   build:
-    docker:
-      # specify the version you desire here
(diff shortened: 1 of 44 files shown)


ksaaskil:
Whoa did not realize `lerna bump` updated the tags: I guess if you now did other commits on this branch before the merge, the tags would be pointing to the wrong version? It's not a big thing, just a minor annoyance I think?

idantene:
That's true.

I'll look into the various options lerna has, I would still maybe prefer to
push the tag during CD.

idantene:
I actually want to also add a changelog in this PR, to get in the habit of
it.

I'd like to also enforce a change to changelog and a version bump in CD,
failing if either is missing.

PR #106, opened 2019-07-25
Update CircleCI

idantene:
Automatic tagging still failed due to missing user name and email definitions.
Adds these and use the native `lerna publish` instead of an npm command.


ksaaskil:
This was cherry-picked to master.

PR #110, opened 2019-07-29
Read env variables from context in the publish step.

ksaaskil:
- [x]  Add required tokens to the context.
- [ ]  Delete tokens from the job's environment variables

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,6 +84,7 @@ workflows:
       - build
       - verify-before-publish
       - publish:
+          context: unmock-publish-nodejs
           requires:
             - build
             - verify-before-publish


idantene:
I think, if we're using a Git token, then we can also remove the credentials helper bash script (and corresponding username/password environment variables).

ksaaskil:
You mean setting `git config user.password $GIT_TOKEN`? This only changes how CircleCI injects $GIT_TOKEN so I'm not sure I understand what you mean by "I think, if we're using a Git token, ..."

idantene:
Yup, or as needed. Just saying I think we can remove the bash script altogether at this point IMO.

ksaaskil:
I kinda think using `git config credential.helper ...` is better, using `git config user.password $GIT_TOKEN` seems more suspectible to leaking the token if someone, for example, wanted to debug the script and set `set -x` in CircleCI 🤔

PR #111, opened 2019-07-29
Throw better error message when no match is found.

ksaaskil:
- Adds a more descriptive error message when no match is found. One could go much fancier here by checking if the `__unmock__` directory exists or if the service existed but I think that can be done when adding CLI stuff.
- Do we want to link docs at some point? Probably not yet as they're not that stable and https://docs.unmock.io even points to wrong docs.

### DIFF
--- a/packages/unmock-node/src/__tests__/backend/index.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/index.test.ts
@@ -45,7 +45,7 @@ describe("Node.js interceptor", () => {
       try {
         await axios("http://example.org");
       } catch (err) {
-        expect(err.message).toBe("No matching template found");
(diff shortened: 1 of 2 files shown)


packages/unmock-node/src/__tests__/backend/index.test.ts
```diff
@@ -45,7 +45,7 @@ describe("Node.js interceptor", () => {
       try {
         await axios("http://example.org");
       } catch (err) {
-        expect(err.message).toBe("No matching template found");
+        expect(err.message).toEqual(expect.stringContaining("No matching template"));
```
idantene: `expect(...).toContain(...)`..?

PR #116, opened 2019-07-30
Refactor UnmockOptions away

idantene:
As `UnmockOptions` was thinned down, it makes little sense to keep it. Instead, _whitelisting_ (as well as `useInProduction` and logging functionalities) have been moved to `CorePackage` layout.
The idea is that one could then change various relevant settings (after initialization) with e.g. `unmock.OPTION = ...`.

This also sets the foundation for other settings as e.g. `unmock.flaky()`.


packages/unmock-core/src/interfaces.ts
```diff
@@ -1,8 +1,7 @@
-import { UnmockOptions } from "./options";
 import { IService } from "./service/interfaces";
 
 export interface ILogger {
-  log: (message: string) => void;
+  log(message: string): void;
```
idantene: Touched these so that IDE-wise they show up as methods and not "callable properties".

packages/unmock-core/src/index.ts
```diff
@@ -1,34 +1,91 @@
...
-    this.on(maybeOptions);
+  public initialize() {
+    this.on();
   }
 
   public off() {
     this.backend.reset();
   }
 
+  public get whitelist() {
```
ksaaskil: Is `whitelist` the best name for this? Maybe something like `allowedHosts` or `passthroughHosts` would make the intention clearer 🤔

packages/unmock-core/src/index.ts
```diff
@@ -1,34 +1,91 @@
...
+    return this.whitelistAsStrings.map((url: string | RegExp) =>
+      url instanceof RegExp ? url.source : url,
+    );
+  }
+  public set whitelist(urls: Array<string | RegExp>) {
```
ksaaskil: Is there anyway to add a URL to allowed hosts in addition to localhost without explicitly re-setting that?

idantene: `unmock.whitelist = unmock.whitelist.concat(["*foo*bar*com"])`?

ksaaskil: Ah ok, so it works by expecting users to set properties instead of exposing methods. Perhaps not the most expected but works 👍

idantene: We can expose it as methods, I have no preference.

ksaaskil: It's fine as is, just thought I'd share the first thought of not expecting to be directly mutating properties for the object in order to change its behaviour.

PR #118, opened 2019-07-30
Make flaky mode the non-default, and set non-existing codes using 'default' response

idantene:
- Move getter/setter to nested method calls as `unmock.allowedHosts.set` and `unmock.allowedHosts.add`, as Kimmo suggested in #116 and later here.
- Enable `unmock.flaky.on()` and `unmock.flaky.off()` to control which response codes are being used. Similarly, `unmock.useInProduction.on()` and `unmock.useInProduction.off()`
- In non-flaky mode, heuristic is as follows: If only one response code exists, use that one; otherwise, if only a single 2XX code is available, used that one; otherwise - throw (suggest using flaky or setting a status code).
- This goes hand in hand with a small modification per OAS (although, maybe we'd prefer to have it as the "old" way) - where a non-existing status code uses the default response. Hence, setting a status code `400` where it doesn't exist, should return the schema specified under `default`.

### DIFF
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -248,11 +248,7 @@ const generateMockFromTemplate = (
       isFlaky: options.flaky(),
     });
 
-  // At this point, we assume there are no references, and we only need to
-  // handle x-unmock-* within the schemas, modify it according to these
-  // properties + the state -> we can work with jsf out of the box
-
-  // Setup the unmock properties for jsf parsing
+  // Setup the unmock properties for jsf parsing of x-unmock-*
   setupJSFUnmockProperties();
   // Always generate all fields for now
   jsf.option("alwaysFakeOptionals", true);
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -181,7 +181,7 @@ const chooseResponseFromOperation = (
   genOptions: { isFlaky: boolean },
 ): {
   $code: string;
-  template: Schema;
+  template: Schema | undefined;
   headers: Headers | undefined;
 } => {
   const responses = operation.responses;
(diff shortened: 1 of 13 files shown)


packages/unmock-core/src/generator.ts
```diff
@@ -87,13 +108,27 @@ const getStateForOperation = (
...
   );
-  // if only one response code is set - use that one; otherwise draw at random
-  // TODO - do we want to set sensible defaults?
-  const statusCode = firstOrRandomOrUndefined(possibleResponseCodes);
-  if (statusCode === undefined) {
+  if (possibleResponseCodes.length === 0) {
     return undefined;
   }
 
+  const statusCode = chooseResponseCode(
```
ksaaskil: Could also do `const statusCode = isFlaky ? ... : ...` to avoid the boolean argument.

packages/unmock-core/src/index.ts
```diff
@@ -71,21 +63,43 @@ export abstract class CorePackage implements IUnmockPackage {
     this.backend.reset();
   }
 
-  public get allowedHosts() {
+  // Allowd Hosts methods
+  public setAllowedHosts(urls: Array<string | RegExp> | string | RegExp): void {
```
ksaaskil: Could also wrap these in a `allowedHosts` object, with `allowedHosts.add` and other methods to reduce the clutter at top-level.

packages/unmock-core/src/index.ts
```diff
@@ -71,21 +63,43 @@ export abstract class CorePackage implements IUnmockPackage {
...
+  public nonFlaky() {
+    this.isFlaky = false;
+  }
+
+  // Use in production methods
+  public useInProduction() {
+    this.activeInProduction = true;
+  }
+  public useInDevelopment() {
+    this.activeInProduction = false;
```
ksaaskil: Not sure if `useInDevelopment` directly translates "do not use in production", to me that's more like "only use in development" 🤔  Maybe could also expose setting objects like `setting.on()` and `setting.off()` i.e. `unmock.someSetting.on()` but it does not really matter.

PR #120, opened 2019-07-31
Move resources in unmock-node tests

idantene:
As the `resources` folder is used across multiple tests (and not only the ones under `loaders`), I thought it makes sense to move it to a more generalized location.
Also adds the slack specification (so we can add more future tests with ease), and a non-parsed petstore specification (the only one had `$ref` resolved manually). I added a `TextError` reference to it so we can have a `text/plain` response as well.

### DIFF
--- a/packages/unmock-node/src/__tests__/backend/index.test.ts
+++ b/packages/unmock-node/src/__tests__/backend/index.test.ts
@@ -2,7 +2,7 @@ import axios from "axios";
 import path from "path";
 import NodeBackend from "../../backend";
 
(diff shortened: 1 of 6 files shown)


idantene:
> Looks good  It could get a bit confusing if we start modifying `petstore` etc. for testing purposes by adding types as it's not _the_ petstore anymore but don't know if it matters.

I generally agree, I just wanted to avoid writing our own made-up specifications at this point. Ideally we will use other specifications for these purposes - we could add another specification that has `text/plain` in the future, to make sure `petstore` is indeed _the_ `petstore` :)

PR #121, opened 2019-07-31
[Fix] Default state without code

idantene:
- Several places used `some` with the underlying assumptions that it iterates over all array elements. Since it does not, some possible states and media types were overlooked once a match was found. Fixed by introducing `anyFn` that acts like `some` but exhausts all array elements before returning.
- Adds many more debug logs.
- Adds typing to accept strings as states by default.
- Adds `test:debug` command (useful for testing a single file).
- Adds a test to verify the above logic, using filestack pseudo API (thanks @ksaaskil!)

### DIFF
--- a/packages/unmock-core/src/service/state/utils.ts
+++ b/packages/unmock-core/src/service/state/utils.ts
@@ -27,22 +27,6 @@ interface ICodesToMediaTypes {
 const debugLog = debug("unmock:state:utils");
 
 /**
- * Works like array.some(), except it exhauses all array elements before returning.
- * @param arr
- * @param callbackfn
- */
-export const anyFn = <T>(
-  arr: T[],
-  callbackfn: (value: T, index: number, array: T[]) => boolean,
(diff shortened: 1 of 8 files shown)


packages/unmock-core/src/service/state/utils.ts
```diff
@@ -26,6 +26,22 @@ interface ICodesToMediaTypes {
 
 const debugLog = debug("unmock:state:utils");
 
+/**
+ * Works like array.some(), except it exhauses all array elements before returning.
+ * @param arr
+ * @param callbackfn
+ */
+export const anyFn = <T>(
```
ksaaskil: I would not expect the internal implementation of `anyFn` to affect the behaviour unless the `callbackFn` was doing side-effects behind the scenes. I guess this function is made for applying side-effects like that? I think it might be clearer to apply side-effects in a `forEach` or `applyAll` rather than `anyFn` 🤔

PR #122, opened 2019-07-31
Log requests and their responses

idantene:
Maintain a log of requests and the generated responses, so that debugging can be done in hindsight if needed.
Overwrite log file if it is over 5MB at time of instantiation.

Also refactors the `unmock.X` settings to their own submodule to keep things a bit cleaner.


packages/unmock-core/src/generator.ts
```diff
@@ -51,10 +55,33 @@ export function responseCreatorFactory({
...
   const stateStore = stateFacadeFactory(serviceStore);
+  const toIndentedYaml = (
+    input: any, // Adds one more indentation to each line
+  ) =>
+    yaml
+      .dump(input.body ? { ...input, body: JSON.parse(input.body) } : input)
+      .split("\n")
+      .map(
+        (line: string) =>
+          `\t${line.replace("!<tag:yaml.org,2002:js/undefined> ''", "")}`,
```
idantene: Removes `js/undefined` tag; maybe we want to keep it? Or maybe we want to use JSON format? I felt yaml is more readable as a log.

packages/unmock-core/src/generator.ts
```diff
@@ -51,10 +55,33 @@ export function responseCreatorFactory({
...
+      )
+      .join("\n");
   return {
     stateStore,
-    createResponse: (sreq: ISerializedRequest) =>
-      generateMockFromTemplate(options, serviceStore.match(sreq)),
+    createResponse: (sreq: ISerializedRequest) => {
+      const resp = generateMockFromTemplate(options, serviceStore.match(sreq));
+      if (logger) {
+        logger.log(
```
idantene: A bit ugly, but rather do one IO operation here.

ksaaskil: This is fine, maybe for future flexibility I'd consider naming the `logger` here as some kind of `listeners` array and do `listeners.forEach.(listener => listener.notify({ req: sreq, res })`)` so that one could, for example, flexibly add the intercepted requests and responses to logs, write to files, and/or add to the list of intercepted request-response pairs?

idantene: Sure, and then accessing that would be via e.g. `unmock.notifiers.add()`?
The types are bit in the way but I'll refactor it as needed if we decide to go that way

ksaaskil: I don't think we would need to expose that to the user, just changing the `createResponseFactory` to notify all injected listeners. Maybe even adding a `try-catch` here so it can notify both success and failure before returning/throwing. That would just also leads us closer to having the `intercepted` array available at least internally.

ksaaskil: I just don't think it's a `responseCreator`'s responsibility to start formatting lines etc.

idantene: I understand that; I don't see why not expose it to the end-user (allowing them to extend functionality if needed), but I don't really don't mind. Will open a PR shortly.

PR #124, opened 2019-08-01
Accept functions (accepts serialized request) for states + bug fixes

idantene:
Apologies in advance for a bigger-than-desired PR :frowning_face:
If really necessary, I could make 3 different PRs from this one, as it introduces 1 feature and resolves 2 bugs. To make things easier, each has the list of files (excluding test files) that are relevant.

- **Feature**: Accepts functions as states, either as entire state generator or as a value. When used as a state generator, it will also receive the schema (although it does not have to use it). Most importantly, this allows:
  1) Verifying requests and/or creating content dynamically
  2) Overriding default schema (no type checking is done for the result, although we could add that if we decide to?)
  Examples include:
  ```typescript
  // generate content dynamically
  states.someService({foo: { id: (req) => req.body.id } });
  // overrides media-types, schema, etc
  states.someService("/get/user", (req, schema) => "I shall ignore the schema");
  // verify request with expect
  states.someService("/get/users/*", { userid: (req) => { expect(req.body.id).toEqual(12345); return 
  req.body.id; } });
  // use with DSL:
  states.someService("/foo/bar", functionResponse((res, schema) => { /* do stuff */ }, { $code: 200 }));
  ```
  *Files*: `generator.ts` (stuff relating to `setupJSFUnmockProperties`), `serviceStore.ts`, `transformers.ts` (adding `functionResponse` and allowing `function` as a type), `utils.ts` (use default schema if the requested code does not explicitly exist)

- **Bugfix**: When setting a state with status code that does not exist, the *default* pattern should be used, but when accessing that state, it should still respond with the requested status code. Fixed this.
  *Files*: `generator.ts`

- **Bugfix**: When using `$size` in some cases it would either not recognize the schema array, or output a wrong error message. This was fixed, and also initiated some much needed refactoring (specifically, how to choose which error message are most relevant).
  *Files*: `dsl/interfaces.ts` (describe DSL keys), `service/interfaces.ts` (refactor to include an error response), `state/interfaces.ts` (refactor to include `IValidationError`), `state.ts` (small refactor to use loop instead of `forEach`, and adapt to new error handling), `transformers.ts` (adapt to return possible error alongside parsed schema, and move some responsibility from `validator.ts`), `utils.ts` (simplified), `validator.ts` (restructure, simplify)

### DIFF
--- a/packages/unmock-core/src/service/state/validator.ts
+++ b/packages/unmock-core/src/service/state/validator.ts
@@ -191,7 +191,7 @@ const getStateFromMedia = (
   const errors: IMissingParam[] = [];
   const relevantResponses: IResponsesFromContent = {};
   let success = false;
-  Object.keys(contentRecord).forEach((contentType: string) => {
+  for (const contentType of Object.keys(contentRecord)) {
     const content = contentRecord[contentType];
     if (content === undefined || content.schema === undefined) {
       debugLog(`getStateFromMedia: No schema defined in ${contentType}`);
@@ -199,9 +199,13 @@ const getStateFromMedia = (
         msg: `No schema defined in '${JSON.stringify(content)}'!`,
         nestedLevel: -1,
       });
-      return;
+      continue;
     }
-    const spreadState = state.gen(deref<Schema>(content.schema));
+    const { spreadState, error } = state.gen(deref<Schema>(content.schema));
+    if (error !== undefined) {
+      errors.push({ msg: error, nestedLevel: 0 });
+      continue;
+    }
 
     debugLog(
       `getStateFromMedia: Copied matching state, verifying all state elements exist (not null)`,
@@ -215,12 +219,12 @@ const getStateFromMedia = (
           `${contentType} is invalid - ${spreadState}`,
       );
       errors.push(missingParam);
-      return;
+      continue;
     }
     debugLog(`getStateFromMedia: Spread state is valid for ${contentType}`);
     relevantResponses[contentType] = spreadState;
     success = true;
-  });
+  }
   return {
     responses: relevantResponses,
     error: success ? undefined : chooseDeepestMissingParam(errors),
--- a/packages/unmock-core/src/service/state/validator.ts
+++ b/packages/unmock-core/src/service/state/validator.ts
@@ -14,6 +14,8 @@ import {
   Responses,
   Schema,
 } from "../interfaces";
+import { IValidationError } from "./interfaces";
+import { chooseBestMatchingError } from "./utils";
 
 const debugLog = debug("unmock:state:validator");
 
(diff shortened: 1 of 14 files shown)


packages/unmock-core/src/service/state/validator.ts
```diff
@@ -134,124 +163,54 @@ const validStatesForStateWithoutCode = (
...
 } => {
   debugLog(
     `getStateFromMedia: Attempting to copy a partial state for ${state} from given media types ${contentRecord}`,
   );
-  const errors: IMissingParam[] = [];
+  let err: IValidationError | undefined;
   const relevantResponses: IResponsesFromContent = {};
   let success = false;
-  Object.keys(contentRecord).forEach((contentType: string) => {
+  for (const contentType of Object.keys(contentRecord)) {
```
ksaaskil: Personally I would some day like to get rid of these big for-loops that may be or may not have side effects, they are quite hard for me to follow but it doesn't matter now.

idantene: It mostly looks big because of line breaks, debug logs and type checks, unfortunately. I don't like the use of `forEach` in these contexts.
There are no side effects either, a result is saved either to `error` or to a possible state.

When the day comes, I'd be happy to learn and see how you'd refactor this though to avoid a (for) loop :)

ksaaskil:
By side-effects I mean setting err, success and modifying a map from inside
the loop. Personally I just think stuff just like map, reduce, filter and
filter and easier to reason about when working on loops.

idantene:
> ... So does this now allow modifying the response with `github.post("*", (req, res) => res)`, I guess not if the second argument is schema? What's the use case for that?

It does! A `res` parameter is redundant is we're creating the response from the state. So the function is expected to return a state of some sort (it will not be validated against the original schema though). That's why the schema is passed - so you can check against it, populate as you'd like, etc.
One could just as well do `github.post((req) => "do stuff")` - you don't have to use the schema. You also don't have to use the request of course.
The idea is that the state for a matching endpoint will be created dynamically, and `unmock` still takes care of things behind the scenes, freeing the user from handling request/response etc.

"Live" example:
```typescript
beforeAll(() => unmock.on());
afterAll(() => unmock.off());
beforeEach(() => unmock.states().reset());

test("foo", async () => {
  unmock().states().petstore("/pets", () => "Hello World!");
  expect((await axios("http://petstore.swagger.io/v1/pets")).data).toEqual("Hello World");
  expect((await axios("http://petstore.swagger.io/v1/pets/3")).data).not.toEqual("Hello World");
});
```

idantene:
> By side-effects I mean setting err, success and modifying a map from inside the loop. Personally I just think stuff just like map, reduce, filter and filter and easier to reason about when working on loops.

Then I really don't understand your intent :)
When this was done earlier with e.g. `reduce`, you gave the same argument that it doesn't make sense to have side-effects like that in `reduce`. I'll see if I can shorten it still.

ksaaskil:
Interesting, I thought we would let users modify the response that would be
returned, either augmenting that or replacing completely. That would make
it easier to return valid responses from the state, it would be a kind of a
middleware.

packages/unmock-core/src/service/state/validator.ts
idantene: Do you prefer the following, for example? (with setting `error` to `undefined` elsewhere -- in `validStatesForStateWithCode`)

```typescript
const getStateFromMedia = (
  contentRecord: Record<string, MediaType>,
  state: IStateInputGenerator,
  deref: Dereferencer,
): {
  responses: IResponsesFromContent;
  error?: IValidationError;
} => {
  debugLog(
    `getStateFromMedia: Attempting to copy a partial state for ${state} from given media types ${contentRecord}`,
  );
  return Object.entries(contentRecord).reduce(
    (
      acc: { responses: IResponsesFromContent; error?: IValidationError },
      [contentType, content],
    ) => {
      if (content == undefined || content.schema === undefined) {
        debugLog(`getStateFromMedia: No schema defined in ${contentType}`);
        acc.error = chooseBestMatchingError(
          {
            msg: `No schema defined in '${JSON.stringify(content)}'!`,
            nestedLevel: -1,
          },
          acc.error,
        );
        return acc;
      }
      const gen = state.gen(deref<Schema>(content.schema));
      if (gen.error !== undefined) {
        acc.error = chooseBestMatchingError(
          { msg: gen.error, nestedLevel: 0 },
          acc.error,
        );
        return acc;
      }
      debugLog(
        `getStateFromMedia: Copied matching state, verifying all state elements exist (not null)`,
      );

      debugLog(`getStateFromMedia: Spread state is valid for ${contentType}`);
      acc.responses[contentType] = gen.spreadState;
      return acc;
    },
    { responses: {} },
  );
};
```

ksaaskil:
> > By side-effects I mean setting err, success and modifying a map from inside the loop. Personally I just think stuff just like map, reduce, filter and filter and easier to reason about when working on loops.
> 
> Then I really don't understand your intent :)
> When this was done earlier with e.g. `reduce`, you gave the same argument that it doesn't make sense to have side-effects like that in `reduce`. I'll see if I can shorten it still.

Earlier it was doing a side-effect in `anyFn` and [I commented](https://github.com/unmock/unmock-js/pull/121/files/9f3518c6671eacf031abd414a50ed8141b3500a4#r309216851) that I found it to be confusing. Having a side-effect in a `reduce` would be equally confusing imho. `forEach` and `for` are made for side effects (`forEach` even more so) so it's clear that there's some mutation going on inside them, but still I think it's always better to avoid `let` wherever possible. Of course it's a question of whether one prefers
```js
let sum = 0;
for (let i = 0; i < arr.length; i++) {
  sum += arr[i];
}
```
vs.
```js
const sum = arr.reduce((a, b) => a+ b, 0);
```
Both are correct.

idantene:
But... you just commented saying `... for-loops that may be or may not have side effects`, followed by  `stuff just like map, reduce, and filter are easier to reason about when working on loops` (I allowed myself to fix typos here).

I understand the preference to avoid `let`, that's fine. I'd also understand if you said you think using `X` over `Y` is more readable. I just don't follow your arguments on side effects here.

ksaaskil:
Ok, let's then just settle with avoiding the use of `let` when possible :)

PR #129, opened 2019-08-05
Deserialize application/json body and do not force it to empty string when missing

ksaaskil:
- This ensures that `req.body` is an object when setting the state with a function with `states.github((req) => { // req.body is an object and not string }`
- Deserializes the body for `application/json` content type
- Do not force the body to be an empty string for empty body
- Do not `JSON.parse` body in the logging phase
- Add a test for `post` request


ksaaskil:
It does not, the only place where the "serialized request" is exposed to the user is if he's trying to use it for setting the state like `states.github.post("**", (req) => { // Do something with the req }`.

PR #134, opened 2019-08-06
Use Babel to transpile to js.

ksaaskil:
- Should make unmock-js work in Node 6
- Use Babel to transpile TS to JS with `@babel/typescript` preset
- Use `tsc` to only build type declarations (and check types)
- Use [core-js@3](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md) to add the polyfills required by the target (Node.js 6) as instructed also [here](https://stackoverflow.com/a/53142387/10561443)

This ensures that for examples usage of `Object.entries()` gets transpiled to a file with
```
require("core-js/modules/es.object.entries");
...
Object.entries(...)
```
so `Object.entries` is polyfilled.
- Keep using ts-jest as the coverage seemed to break when relying on babel-jest to transform TS files (might require regenerator runtime?)

Resources:
- [TypeScript + Babel: A Beautiful Marriage](https://iamturns.com/typescript-babel/)
- [Microsoft/TypeScript-Babel-Starter](https://github.com/Microsoft/TypeScript-Babel-Starter)
- [Monorepo Setup With TS + Babel + Lerna](https://medium.com/grand-parade/monorepo-setup-with-lerna-typescript-babel-7-and-other-part-1-ac60eeccba5f?) (see the [repo](https://github.com/serhii-havrylenko/monorepo-babel-ts-lerna-starter))
- [Jest](https://github.com/facebook/jest) uses TypeScript with Babel


jest.config.base.js
```diff
@@ -11,8 +11,14 @@ module.exports = {
     "^(unmock|unmock-cli|unmock-core|unmock-jsdom|unmock-node)(?:/dist)?((?:/.*)|$)":
       "<rootDir>/../../packages/$1/src$2"
   },
-  setupFilesAfterEnv: ["<rootDir>/../../jest.setup.js"],
   clearMocks: true,
+  transformIgnorePatterns: [
+      "/node_modules/(?!loas3)",
```
ksaaskil: Required in Node 6 as loas does not transpile and uses stuff like `Object.entries`.

package.json
```diff
@@ -2,12 +2,14 @@
   "license": "MIT",
   "scripts": {
     "clean": "npm prune && del-cli ./packages/*/dist ./packages/*/tsconfig.tsbuildinfo",
-    "compile": "tsc --build tsconfig.build.json",
-    "compile:clean": "tsc --build tsconfig.build.json --clean",
-    "compile:tests": "tsc --build tsconfig.test.json && tsc --build tsconfig.test.json --clean",
+    "build": "npm run build:types:clean && npm run check:types:tests && npm run build:types && npm run build:js",
```
ksaaskil: For some reason building the type declarations *after* building JS fails, needs to build type declarations first.

PR #170, opened 2019-08-19
Make DSL submodule more FP

idantene:
Following the conversation in #124 (and out of my own personal wishes to improve my skills in FP :innocent: ), I thought to slowly transition stuff to have more FP-style.
Main goal is to find `for (` in the code and replace them as possible with the matching functional methods.

* `dsl.test.ts` modified because `DSL.translateDSLToOAS` was renamed to `DSL.replaceDSLWithOAS` to better reflect underlying processes.

### DIFF
--- a/packages/unmock-core/src/service/dsl/actors.ts
+++ b/packages/unmock-core/src/service/dsl/actors.ts
@@ -1,25 +1,29 @@
 import debug from "debug";
+import { cloneDeep } from "lodash";
 import { mediaTypeToSchema } from "../interfaces";
 import { SCHEMA_TIMES } from "./constants";
-import { Props } from "./interfaces";
+import { Actor, Props } from "./interfaces";
 
 const debugLog = debug("unmock:dsl:actors");
 
 /*
- * Actors (actOn$X) are found here. They modify the behaviour/schemas according to specific DSL instructions.
+ * Actors (actOn$X) are found here.
+ * They modify the behaviour/schemas according to specific DSL instructions.
+ * They all use the Actor type.
  */
 
 /**
- * Acts on the $times argument in top level DSL by synchronizing and modifying the copied and original schemas.
+ * Acts on the $times argument in top level DSL by MODIFYING the original
+ * schema as needed and returning a relevant copy.
+ *
  * The value of $times is decreased by 1 and removed from the copied schema.
  * If the new value is less than 0 (i.e. this state has expired),
- * the content is removed from **both** the copied and original schemas.
- * @param copiedSchema
+ * the content is **removed** from the original schema, and the returned object is empty.
+ *
  * @param originalSchema
  * @param mediaType
  */
-export const actOn$times = (
-  copiedSchema: mediaTypeToSchema,
+const actOn$times: Actor = (
   originalSchema: mediaTypeToSchema,
   mediaType: string,
 ) => {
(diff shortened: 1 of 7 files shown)


packages/unmock-core/src/service/dsl/actors.ts
```diff
@@ -1,25 +1,29 @@
 import debug from "debug";
```
mikesol: If you're looking for a good FP start, this is a great function to do it in - it can be rewritten to create a new object with only the `.default` parameter changed and will probably be more performant than the deep copy (although we are talking about single-digit microseconds in the grand scheme of things, so the only real reason to do it would be to have fun with FP).

packages/unmock-core/src/service/dsl/dsl.ts
```diff
@@ -33,16 +28,20 @@ export abstract class DSL {
...
-  public static translateDSLToOAS(state: any, schema: Schema): any {
-    let translated: { [OASKey: string]: string | number | boolean } = {};
-    if (state.$size !== undefined) {
-      throwOnErrorIfStrict(() => {
-        translated = { ...translated, ...translate$size(state, schema) };
-        delete state.$size;
-      });
-    }
-    return translated;
+  public static replaceDSLWithOAS(state: any, schema: Schema): any {
```
mikesol: Ditto, easy candidate for removing the `delete`.

packages/unmock-core/src/service/dsl/dsl.ts
```diff
@@ -59,12 +58,11 @@ export abstract class DSL {
     }
     // Handles top-level schema and injects the literals to responses.
     // $code is a special case, handled outside this function (acts as a key and not a value)
-    if (top.$times !== undefined) {
-      throwOnErrorIfStrict(() => {
-        const translated = translate$times(top.$times);
-        injectUnmockProperty(responses, translated);
-      });
-    }
+    Object.keys(topTranslators)
```
mikesol: Great! One common pattern (that I am guilty of being too lazy to do) is the following:

```ts 
    Object.keys(topTranslators) 
      .filter(dslKey => top[dslKey] !== undefined /* no undefined tops */) 
      .map(dslKey => topTranslators[dslKey](top[dslKey]) /*gt the translation */) 
      .filter(translation => translation !== undefined /* no undefined translation */) 
```

etc...

packages/unmock-core/src/service/dsl/actors.ts
idantene: Could you elaborate? The`default` is not needed in the copied version (which is later used to generate the response). However, the DSL does not interact with a `State` object immediately, so it modifies a passing state as needed (reduces the `default` value for the `SCHEMA_TIMES` property).

Could remove the `delete` statement if I could copy anything *but* that specific key. It seems easier to do a deep copy; do you have any suggestions?

idantene: `const {x, ...y} = {x: 5, a: 2, b: 3 }` works only with hard coded names, otherwise that would've been ideal :thinking:

idantene: Ah, managed to achieve this with some destructuring (removing the deep copy and the last `delete` statement).

There is still a `delete` statement to modify the `State`'s state as needed. I think that's fine for now (at least the topic for a different PR, as it would require bigger changes).

ksaaskil: Not sure if it helps but lodash has `_.omit` and `_.omitBy` that could help avoid the deletes.

packages/unmock-core/src/service/dsl/dsl.ts
ksaaskil: This function is a bit hard for me to understand 🤔 It's modifying the input object by deleting properties from it and at the same time it's also returning stuff. I think it would be better to have either-or: either only modify the input and return `void`, or make it a pure function returning a new clone of the object, possibly along with the list of translated arguments if that's needed.

ksaaskil:
And personally I would not worry about performance unless we're deep-copying humongous objects like OpenAPI schemas (cannot see if we are, don't know why we should be). But if performance then becomes a bottleneck, we can consider using `immutable-js` (or just go back to mutating stuff).

packages/unmock-core/src/service/dsl/dsl.ts
idantene: Yeah that's fair. Will look into it (require changes elsewhere actually, but could be a code smell).

idantene: Will be a different PR though.

PR #171, opened 2019-08-19
Groundwork for a service call spy

ksaaskil:
- Create a `CallTracker` that one should be able to attach to a service
- `callTracker.track({ req, res })` indirectly adds spied calls to `callTracker.spy`
- This should work until services have methods to spy on themselves, if we decide to go that way

### DIFF
--- a/packages/unmock-core/src/service/spy/index.ts
+++ b/packages/unmock-core/src/service/spy/index.ts
@@ -29,44 +29,33 @@ export type RequestResponseSpyNotifier = CallTracker<
 >;
 
 /**
- * Helper class whose `notify(arg, ret)` calls `track` with argument `arg`
- * with predefined response `ret`. This is useful when you want to "spy" on `track`
- * with predefined request-response pairs.
- */
-export class SpyHelper<TArg = any, TReturnValue = any> {
-  private returnValue?: TReturnValue;
-
-  public notify(targs: TArg, tret: TReturnValue) {
-    // Ugly hack to predefine what the spied function should return
-    this.returnValue = tret;
-    this.track(targs);
-    this.returnValue = undefined;
-  }
-
-  public track(_: TArg): TReturnValue {
-    return this.returnValue!;
-  }
-}
-
-/**
  * Container for tracking spy calls via `track` method,
  * calls are exposed via `spy` property.
  */
 class CallTracker<TArg = any, TReturnValue = any> {
   // tslint:disable-line:max-classes-per-file
   public readonly spy: SinonSpy<[TArg], TReturnValue>;
-  private readonly recorderHelper: SpyHelper<TArg, TReturnValue>;
+  private returnValue?: TReturnValue;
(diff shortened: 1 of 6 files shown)


packages/unmock-core/src/service/spy/index.ts
```diff
@@ -0,0 +1,63 @@
+import { SinonSpy, spy as sinonSpy } from "sinon";
```
idantene: This is a bit confusing (having both `sinonSpy` and `SinonSpy`). How about `import { SinonSpy as SinonSpyType, ...`?

packages/unmock-core/src/service/spy/index.ts
```diff
@@ -0,0 +1,63 @@
...
+   */
+  spy: RequestResponseSpy;
+  /**
+   * Add new call to the spy object
+   */
+  track(pair: IRequestResponsePair): void;
+  reset(): void;
+}
+
+export type RequestResponseSpyNotifier = CallTracker<
```
idantene: Why are the above interfaces/types exported? :thinking:

packages/unmock-core/src/service/spy/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  ISerializedRequest,
+  ISerializedResponse
+>;
+
+/**
+ * Container for tracking spy calls via `track` method,
+ * calls are exposed via `spy` property.
+ */
+class CallTracker<TArg = any, TReturnValue = any> {
```
idantene: `... implements ICallTracker`? (not 100% convinced an interface is needed here, especially if we'll move this to be under the `Service` class)

packages/unmock-core/src/service/spy/index.ts
idantene: Also this specific type (`RequestResponseSpyNotifier`) is not used anywhere at the moment.

ksaaskil:
Yep we're spying indirectly because there's no method on a service instance to spy. But that's not something users need to care about in my opinion.

packages/unmock-core/src/service/spy/index.ts
ksaaskil: I just think it's cleaner to export an interface instead of concrete class.

ksaaskil: And yes that `RequestResponseSpyNotifier` was a left-over, removed.

packages/unmock-core/src/service/spy/index.ts
ksaaskil: Could implement if it wasn't generic. Maybe it would be cleaner to have the types be concrete but they don't really need to be so thought it could be generic.

packages/unmock-core/src/service/spy/index.ts
ksaaskil: Renamed.

PR #173, opened 2019-08-20
Functional DSL Part 2

idantene:
- Move state management to be solely in `State` -> actors and DSL do not change state but rather return a copy of "state to return" and possibly updated state. `undefined` is used to denote a "state-to-be-delete".

- Removes the idea that multiple states apply to a single endpoint (e.g. previously if `/pets/*` was set with `{ type: "dog" }` and `/pets/3` was set to `{ name: "sparkie" }`, then hitting `/pets/3` would return `{ name: "sparkie", type: "dog" }`). This is still nice IMO, but might not the expected result for many users. Moreover, it gets messier when using a DSL keyword, and most importantly, it proves quite hacky when updating a state based on the DSL response. So now, only the most relevant endpoint is chosen.

### DIFF
--- a/packages/unmock-core/src/service/state/state.ts
+++ b/packages/unmock-core/src/service/state/state.ts
@@ -152,9 +152,10 @@ export class State {
 
     // Filter all the states that do not match the operation schema
     const filteredStates = filterStatesByOperation(states, operation);
-    const parsedTopLevel = DSL.actTopLevelFromOAS(filteredStates);
+    const { parsed, newState } = DSL.actTopLevelFromOAS(filteredStates);
+    console.log(newState);
     // TODO: After parsing, do we want to see if we need to remove items from the current state?
-    return Object.keys(parsedTopLevel).length > 0 ? parsedTopLevel : undefined;
+    return Object.keys(parsed).length > 0 ? parsed : undefined;
   }
 
   public reset() {
--- a/packages/unmock-core/src/service/state/state.ts
+++ b/packages/unmock-core/src/service/state/state.ts
@@ -87,7 +87,7 @@ export class State {
   }
 
   /**
-   * Returns the state for given combination of method and endpoint.
+   * Returns the state for given combination of method and endpoint. Might update state according to DSL.
    * At this point, method is expected to be a valid HTTP method, and endpoint is expected
    * to be specific endpoint.
    * @param method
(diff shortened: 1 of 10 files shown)


packages/unmock-core/src/service/state/state.ts
```diff
@@ -111,50 +102,63 @@ export class State {
...
+          // Marked for deletion
+          this.deleteStateInternal(
+            mostRelevantEndpoint,
+            matchingMethod,
+            code,
+            media,
+          );
+        } else if (Object.keys(state).length > 0) {
+          // new state available
+          this.updateStateInternal(mostRelevantEndpoint, matchingMethod, {
```
ksaaskil: It's a bit surprising to me that a method called `getState` is doing all these updates and deletes to `this.state`, "get" sounds like something pure to me...

idantene: I thought the same. At the same time, I don't think it's `Service`'s responsibility to do these updates.
My intention was that `getState` would actually indicate the return value, and what happens under the hood is abstracted away. Open for any suggestions though.

idantene: At any case, the only time this update even happens (for now, at least), is when `$times` has reached 0, which is directly related to the `getState` operation. I don't see another reason why it would need to be updated after a get operation.

ksaaskil: Hmm, should there then be a simple `if (timesReachedZero(state)) { this.state = updateState(state); }` guard for that case?

idantene:
That's essentially what's being done (with a general case included if we
add other cases where the DSL could affect the state), albeit `this.state`
being mutable (simply because it's either to find what we need to remove,
rather then iterating over what needs to stay) and that can be changed of
course :)

The caveats are that $times reaching 0 is already checked elsewhere (in the
DSL). Even if we remove that, I don't think it's a state's responsibility
to know about the DSL. Neither is it the DSL responsibility to know how
`this.state` is saved.

All in all, I think such a change would end up looking like (I hope this
works via email):

```typescript

Object.keys(newState).forEach(code =>
      Object.keys(newState[code]).forEach(media => {
        const state = newState[code][media];
        if (DSL.hasZeroTimesLeft(state)) {
          this.state = this.removeStateFor(
            mostRelevantEndpoint,
            matchingMethod,
            code,
            media,
          );
        }

      });

);

```

ksaaskil:
Ok cool. Personally I find the nested `forEach` with anonymous functions quite hard to read, I'd maybe add more helper functions so it's easier to understand what's going on.

idantene:
Yeah that's fair, I'm not a fan of that either :/
Could abstract it to a `forEveryCodeMediaTypeCombo` or something less
verbose.

PR #174, opened 2019-08-20
FP leftovers

idantene:
Removes more `for` loops, `let` statements where possible. Simplifies some actions taken in both `validator` and `transformers`.

With this, I feel like the FPying process is done.

### DIFF
--- a/packages/unmock-core/src/service/serviceStore.ts
+++ b/packages/unmock-core/src/service/serviceStore.ts
@@ -25,20 +25,16 @@ export class ServiceStore {
   }
 
   public match(sreq: ISerializedRequest): MatcherResponse {
-    for (const service of Object.values(this.serviceMapping)) {
-      const matchResponse = service.match(sreq);
-      if (matchResponse !== undefined) {
-        return matchResponse;
-      }
-    }
-    return undefined;
+    return Object.values(this.serviceMapping)
+      .map(service => service.match(sreq))
+      .filter(res => res !== undefined)[0];
   }
 
   public resetState(serviceName: string | undefined) {
     if (serviceName === undefined) {
-      for (const service of Object.values(this.serviceMapping)) {
-        service.resetState();
-      }
+      Object.values(this.serviceMapping).forEach(service =>
+        service.resetState(),
+      );
     } else if (this.serviceMapping[serviceName] !== undefined) {
       this.serviceMapping[serviceName].resetState();
     }
--- a/packages/unmock-core/src/service/serviceStore.ts
+++ b/packages/unmock-core/src/service/serviceStore.ts
@@ -25,6 +25,7 @@ export class ServiceStore {
   }
 
   public match(sreq: ISerializedRequest): MatcherResponse {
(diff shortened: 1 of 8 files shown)


packages/unmock-core/src/service/util.ts
```diff
@@ -26,12 +26,18 @@ export function derefIfNeeded({
       return mightHaveReference;
     }
     // DFS deref all items as needed
-    for (const childKey of Object.keys(mightHaveReference)) {
-      mightHaveReference[childKey] = selfDeref(mightHaveReference[childKey]);
-    }
-    const refValue = mightHaveReference.$ref;
+    const noReferences = Array.isArray(mightHaveReference)
```
idantene: Extra code but feels more verbose :roll_eyes:

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -25,20 +25,16 @@ export class ServiceStore {
...
-    for (const service of Object.values(this.serviceMapping)) {
-      const matchResponse = service.match(sreq);
-      if (matchResponse !== undefined) {
-        return matchResponse;
-      }
-    }
-    return undefined;
+    return Object.values(this.serviceMapping)
+      .map(service => service.match(sreq))
+      .filter(res => res !== undefined)[0];
```
ksaaskil: Taking the first element of a possibly empty list looks very dangerous but I guess JS allows that so...For stuff like this the `Option` data type would be perfect.

packages/unmock-core/src/service/state/utils.ts
```diff
@@ -173,3 +170,11 @@ export const chooseBestMatchingError = (
     ? secondError
     : firstError;
 };
+
+export const chooseErrorFromList = (
+  errList: Array<IValidationError | undefined>,
```
ksaaskil: And for this kind of stuff `Validation` data type would be perfect.

packages/unmock-core/src/service/state/transformers.ts
```diff
@@ -113,129 +195,53 @@ const spreadStateFromService = (
...
 
 const isConcreteValue = (obj: any) =>
   ["string", "number", "boolean", "function"].includes(typeof obj);
 
-const isNonEmptyObject = (obj: any) =>
-  typeof obj === "object" && Object.keys(obj).length > 0;
+const isEmptyObject = (obj: any) =>
+  typeof obj !== "object" || Object.keys(obj).length === 0;
 
 const oneLevelOfIndirectNestedness = (
```
ksaaskil: TBH I have no idea what e.g. this method does :) Do we still need to have this many `any`s?

idantene: The `oneLevelOfIndirectNestedness`?

Re `any`, most of them can probably be replaced with a `Schema` object, actually.

idantene: Ah, the reason it was `any` is because `Schema` object does not define a `[key: string]: ...` signature. I'll just cast it to such with an intermediate type.

Issue #150, opened 2019-08-09
Merge unmock-core to unmock-node


ksaaskil:
`unmock-core` was originally an isomorphic package usable both in Node.js and browser. The browser implementation is not needed anymore, so `unmock-core` can be merged to `unmock-node` inside, for example, `core` directory.

Also delete `unmock-jsdom`.

idantene:
Done in #182 (albeit merged `unmock-node` into `unmock-core` instead of the other way around).

PR #175, opened 2019-08-21
Export `sinon` from unmock-core and unmock-node

ksaaskil:
This allows doing 
```js
import { sinon: { assert, match }} from "unmock-node";
assert.wasCalledOnce(github.spy);
assert.wasCalledWithMatch(github.spy, { path: match(/^\/v3/) });
```
I think that's already quite powerful (though not 100% readable) before we have implemented any custom asserts or matchers maybe allowing stuff like `expect(github).calledOnceWith(matchPath(/^\/v3/)`.

Exposing `sinon` could be a temporary solution, but I don't see harm in exposing that. At least not until we have `unmock-expect` implemented.

Alternative would be to expose `import { assert, match } from "unmock-node"` but I'm a bit worried it might turn out a mess: is `unmock.match` then `sinon.match`, superset of `sinon.match`, or a set disjoint of `sinon.match`? I'd maybe reserve `unmock.match` for future for clarity 🤔

### DIFF
--- a/packages/unmock-core/src/index.ts
+++ b/packages/unmock-core/src/index.ts
@@ -14,7 +14,8 @@ export {
   CustomConsole as UnmockConsole,
 } from "./console";
 
-export { default as sinon } from "sinon";
+import sinon from "sinon";
+export { sinon };
 
 import { StateFacadeType } from "./service/interfaces";
 export type States = StateFacadeType;
--- a/packages/unmock-core/src/index.ts
+++ b/packages/unmock-core/src/index.ts
@@ -14,6 +14,8 @@ export {
   CustomConsole as UnmockConsole,
 } from "./console";
 
+export { default as sinon } from "sinon";
+
 import { StateFacadeType } from "./service/interfaces";
 export type States = StateFacadeType;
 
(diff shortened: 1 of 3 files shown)


idantene:
Side bar - all of these recent modifications IMO call for us to unify `unmock-core` with `unmock-node`, remove `unmock-js`, and essentially have our packages include:
- `unmock` for shorthand notation and installation
- `unmock-core` (or `node`, whatever naming we choose for it)
- `unmock-cli`
- `unmock-expect`

ksaaskil:
Yep, there's this issue for merging unmock-core to unmock-node https://github.com/unmock/unmock-js/issues/150

packages/unmock-core/src/index.ts
```diff
@@ -14,6 +14,9 @@ export {
   CustomConsole as UnmockConsole,
 } from "./console";
 
+import sinon from "sinon";
+export { sinon };
```
ksaaskil: `export { default as sinon } from "sinon"` does not work here, not sure why.

idantene:
Ah cool didn't notice it :sweat_smile:

PR #176, opened 2019-08-21
Refactor user facing API

idantene:
Big PR ahead :skull_and_crossbones: 

- **Removes** the `states` object, instead exposes `unmock.services` with `unmock.services.serviceName.state....` as usual
- Renames `Service` to `ServiceCore`, adds a new user facing `Service`
- Updates types and tests as needed
- [x] Cleanup
- [x] Add tests for new `Service`


packages/unmock-core/src/index.ts
```diff
@@ -50,18 +48,20 @@ export abstract class CorePackage implements IUnmockPackage {
       log: (message: string) => this.logger.log(message),
       flaky: () => this.flaky.get(),
     };
-    return this.backend.initialize(opts);
+    this.backend.initialize(opts);
   }
   public init() {
-    return this.on();
+    this.on();
```
ksaaskil: So `unmock.on()` is not supposed to return anything anymore?

packages/unmock-core/src/interfaces.ts
```diff
@@ -22,17 +24,18 @@ export interface IUnmockOptions extends ILogger {
...
   reset(): void;
 }
 
 export interface IUnmockPackage {
   allowedHosts: AllowedHosts;
-  on(): States;
-  init(): States;
-  initialize(): States;
+  services: IServiceStore;
+  on(): void;
```
ksaaskil: Any reason not to return `services` from `unmock.on()`? I kinda liked that even if it broke 
```js
beforeAll(() => unmock.on());
```

Maybe we should at some point distinguish between loading services and turning unmock on, those are quite different tasks and should be doable via different commands (of course we'd still expose a command that does both).

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -1,91 +1,20 @@
...
-  objResponse,
-  textResponse,
-} from "./state/transformers";
+import { IService, IServiceCore, IServiceStore } from "./interfaces";
+import { Service } from "./service";
 
-export class ServiceStore {
-  private readonly serviceMapping: IServiceMapping = {};
+export const ServiceStore = (coreServices: IServiceCore[]) =>
+  new Proxy(
```
ksaaskil: Do we still need proxy if `unmock.services` is simply a `Record<string, Service> & ServiceStoreHelpers`?

ksaaskil: At least for me this type-checks fine:
```ts
interface Service {
  name: string;
}

type ServiceStoreHelpers = {
  findByHost(host: string): Service;
};

type Services = Record<string, Service> & ServiceStoreHelpers;

const service: Service = { name: "github " };

const services: Services = {
  findByHost(host: string) {
    if (host === "api.github.com") {
      return service;
    }
    throw Error("No service named github");
  },
  github: service,
};

const github = services.findByHost("api.github.com");
const githubAlt = services.github;
```

ksaaskil: Of course we may need custom getters to ensure that `services.githu` throws but I think the error message is quite obvious is one then tries to access `services.githu.state`.

ksaaskil:
And a great job on this!

packages/unmock-core/src/interfaces.ts
idantene: No reason at all. I wasn't sure if it's best to return e.g. services or some other element of unmock.
I'll update it.

Also really agree on loading services as a different part, I think that should probably be in the constructor. Belongs in a different PR though.

packages/unmock-core/src/service/serviceStore.ts
idantene: We don't need a proxy here at all, it's a question of do we want an error that reads `cannot read property 'state' of undefined` or `unmock: No service named '....'`.

:thinking: At least I'm not aware of how to throw when trying to access `services.githu`

ksaaskil: I prefer the `cannot read property 'state' of undefined` if it means avoiding proxy :)

idantene: Sure. Personally, I think a more verbose message is beneficial for the end-user (just thinking how long it would take to notice if e.g. `services.gthub` is used by mistake, for example), but we can always revert back to this later on.

At this point then, `ServiceStore` has no functionality - it's just a mapping.

packages/unmock-node/src/types.ts
```diff
@@ -3,5 +3,4 @@ export {
   ISerializedRequest as Request,
   ISerializedRequest as UnmockRequest,
   ISerializedResponse as UnmockResponse,
-  States,
```
ksaaskil: Maybe expose  `Services` then?

idantene: Hm, sure. I don't see pros or cons to exposing it (do we expect users to manually type it when it's returned from `unmock.services` as `ServiceStoreType`?)

ksaaskil: I think the types of all exposed values should be exposed, so users don't need to use `any` if doing 
```
let services: any;
beforeAll(() => {
   services = unmock.on().services;
}
```

ksaaskil: The bravest hackers would of course do something like `let services: typeof unmock.services`.

idantene: Good point.

Also, since `services` is a dictionary, one can safely do `beforeEach(() => unmock.on());` now (I guess it tries to access `unmock.services.then`, but that's just a no-op).

packages/unmock-core/src/service/serviceStore.ts
idantene: `findByHost` left for another PR ofc

PR #177, opened 2019-08-22
Return this from unmock.on, add more type exports.

ksaaskil:
- Export `Service` (interface) and `UnmockNode` (alias for `IUnmockPackage`)
- Exporting types is becoming a bit of a mess: I think `unmock-core` should have `types.ts` for all exposed types and export it as e.g. `types` so that `unmock-node` could simply export everything from there. I can try and work on that.

### DIFF
--- a/packages/unmock-node/src/types.ts
+++ b/packages/unmock-node/src/types.ts
@@ -4,5 +4,6 @@ export {
   ISerializedRequest as UnmockRequest,
   ISerializedResponse as UnmockResponse,
   ServiceStoreType as UnmockServices,
+  IService as UnmockService,
   IService as Service,
 } from "unmock-core";
--- a/packages/unmock-node/src/types.ts
+++ b/packages/unmock-node/src/types.ts
@@ -4,4 +4,5 @@ export {
   ISerializedRequest as UnmockRequest,
   ISerializedResponse as UnmockResponse,
   ServiceStoreType as UnmockServices,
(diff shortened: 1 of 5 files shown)


packages/unmock-node/src/types.ts
```diff
@@ -4,4 +4,5 @@ export {
   ISerializedRequest as UnmockRequest,
   ISerializedResponse as UnmockResponse,
   ServiceStoreType as UnmockServices,
+  IService as Service,
```
idantene: Maybe `UnmockService` or similar?

ksaaskil: Does `Service` collide with something or why `UnmockService`? Or would you like to keep it consistent with `UnmockRequest` and `UnmockResponse`? Not sure where to draw the line then, should everything be prefixed with `UnmockSomething`? It also feels a bit extra to write `const github: UnmockService = ...` instead of `const github: Service = ...` but I can of course change it.

ksaaskil: Exported both so one can choose whichever!

idantene: It does feel a bit extra to write that. It doesn't collide with anything AFAICS, but more so that `Service` is quite general, and does not necessarily imply that it's related to unmock library.

We could think of a different prefix of course. Perhaps, for example, a lowercase `u`? :grin:

PR #180, opened 2019-08-23
Release v0.2.0

ksaaskil:
Breaking changes:
  - `unmock.on()` returns `this` instead of `states` object.
  - Deleted `unmock.states`: state changes should be done via `unmock.services["service"].state`

Features:
  - Exports `UnmockRequest`, `UnmockResponse` and other exposed types
  - Adds service spy for keeping track of service calls
  - Exports `sinon` for easy assertions on service spy

Required changes:
  - Documentation: https://github.com/unmock/unmock.github.io/pull/11
  - Examples: https://github.com/unmock/unmock-examples/pull/13


changelog.md
```diff
@@ -1,5 +1,18 @@
 # CHANGELOG.md
 
+## 0.2.0 (2019-08-23)
+
+Breaking changes:
+
+  - `unmock.on()` returns `this` instead of `states` object.
+  - Deleted `unmock.states`: state changes should be done via `unmock.services["service"].state`
```
idantene: `unmock.services.service.state`? (or do you think that's implied?)

ksaaskil: I just thought using the `[]` accessor made it a bit more clear that the `service` part is not part of the API but essentially a variable.

PR #188, opened 2019-08-26
Cleanup

idantene:
- Removes `IBackend`, instead uses `NodeBackend`
- Remove `IServiceDefLoader`, instead passes around `FsServiceDefLoader`
- Moves `parser` and `FsServiceDefLoader` to top level
- Removes some unnecessary exports from `index`
- Removes anecdotal/single-use interfaces from `service/interfaces`
- Remove `unmock-node` package (leftover package.json...)

Not 100% convinced all of these are necessary, just playing around seeing what can be (re)moved. Comments appreciated.


ksaaskil:
Looks good, maybe hold this for tomorrow.

idantene:
Yup np. In case I won't be available, feel free to merge this.

ksaaskil:
I'll merge this, looks good 👍 

As a general comment for the use of interfaces, I kinda think it's a good idea to abstract away anything where one can (1) see a lot of change happening in future and (2) imagine multiple existing implementations. In the case of `IService` (2) maybe doesn't hold so we don't need interface. On the other hand, for `FsServiceDefLoader` I think they do: it does not matter to the user of how and from where those service defs are read, and one can envision new sources of service defs. Even if we never implemented new loaders, I think abstraction in that case reduced coupling. But it does not really matter for now, can always add back in.

Issue #183, opened 2019-08-23
Body in UnmockResponse is string when should be JSON


ksaaskil:
It's unexpected that `body` in `UnmockResponse` (alias for `ISerializedResponse`) is represented as `string` even when the response is `application/json`. Change it to work as `UnmockRequest` so that the body is `string | object | undefined`.

idantene:
`s/UnmockResponse/ISerializedResponse/g`

Where does this happen? Here it's clearly being stringified... https://github.com/unmock/unmock-js/blob/73cc0003623a9bb6162e1e313c1dff72b7c0d734/packages/unmock-core/src/generator.ts#L305

ksaaskil:
So this would mean
- Changing the body type to `string | object`
- Removing the cast to string in the mock generator
- Casting to string when sending the response from Node.js interceptor.

As reference, also `yesno` [represents the response](https://github.com/FormidableLabs/yesno/blob/master/src/http-serializer.ts#L37) as string or object.

ksaaskil:
After this is done, one can remove the `JSON.parse` from [this example](https://github.com/unmock/unmock-examples/blob/a3a2e1df905610181e4e8691ba2ecdb4db19d4b6/hello-world-javascript/hello.test.js#L50).

ksaaskil:
We'll keep the body as `string` for now to avoid typing it as `any` (typing as union type is very cumbersome for testing).

PR #189, opened 2019-08-26
resolve issue #183: Body in UnmockResponse is string when should be JSON

dangquangdon:
- Adjust the type of body in ISerializedResponse interface to `string | object`

-  remove casting to string in the `generator.ts` (line 305)

- adjust test case in `generator.test.ts`

### DIFF
--- a/packages/unmock-core/src/__tests__/generator.test.ts
+++ b/packages/unmock-core/src/__tests__/generator.test.ts
@@ -116,7 +116,7 @@ describe("Tests generator", () => {
     expect(resp).toBeDefined();
     if (resp !== undefined) {
       expect(resp.statusCode).toEqual(204);
-      expect(resp.body).toEqual('"prefetch"');
+      expect(resp.body).toEqual("prefetch");
     }
 
     resp = createResponse({
@@ -128,7 +128,7 @@ describe("Tests generator", () => {
     expect(resp).toBeDefined();
     if (resp !== undefined) {
       expect(resp.statusCode).toEqual(200);
-      expect(resp.body).toEqual('"prefetch"');
+      expect(resp.body).toEqual("prefetch");
     }
   });
 
@@ -146,9 +146,11 @@ describe("Tests generator", () => {
     });
     expect(resp).toBeDefined();
     if (resp && resp.body !== undefined) {
-      JSON.parse(resp.body).forEach((pet: any) =>
-        expect(pet.id).toEqual("foo"),
-      );
+      if (!(resp.body instanceof Array)) {
+        throw new Error("Response body was not an object but a string");
+      }
+
+      resp.body.forEach((pet: any) => expect(pet.id).toEqual("foo"));
     } else {
       throw new Error("Response body was undefined?");
     }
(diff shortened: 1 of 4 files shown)


dangquangdon:
@mikesol , @idantene , @ksaaskil please review

packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -146,9 +146,11 @@ describe("Tests generator", () => {
     });
     expect(resp).toBeDefined();
     if (resp && resp.body !== undefined) {
-      JSON.parse(resp.body).forEach((pet: any) =>
-        expect(pet.id).toEqual("foo"),
-      );
+      if (!(resp.body instanceof Array)) {
```
mikesol: https://jestjs.io/docs/en/expect.html#tobeinstanceofclass

packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -146,9 +146,11 @@ describe("Tests generator", () => {
...
+        throw new Error("Response body was not an object but a string");
+      }
+
+      resp.body.forEach((pet: any) => expect(pet.id).toEqual("foo"));
```
mikesol: I would roll this into one expectation. Depending on what you are trying to test, this could be `expect(new Set (resp.body.map(v => v.id))).toEqual(new Set(["foo"]))`

idantene: Or `expect(resp.body.every((pet: any) => pet.id === "foo")).toBeTruthy();`

packages/unmock-core/src/generator.ts
```diff
@@ -302,7 +302,7 @@ const generateMockFromTemplate = (
   jsf.option("useDefaultValue", false);
   const resolvedTemplate = jsf.generate(template);
 
-  const body = JSON.stringify(resolvedTemplate);
+  const body = resolvedTemplate;
```
idantene: I think `const body = jsf.generate(template);` would be better (+ removing the `resolvedTemplate` above).
Or then returning `{ body: resolvedTemplate }`, etc, but this assignment is no longer needed :)

packages/unmock-core/src/__tests__/generator.test.ts
ksaaskil: That does not do the type-checking that's necessary for the following `resp.body.forEach`.

packages/unmock-core/src/__tests__/generator.test.ts
ksaaskil: Note that this line was not added in this PR, but of course can update that as well.

packages/unmock-core/src/__tests__/generator.test.ts
idantene: is it better to use e.g. Array.isArray(resp.body) here? :thinking: 
I know it's relatively new (ES5?) but as it's used elsewhere in the code, I don't see the harm.
It might be more readable?

ksaaskil: As long as TypeScript understands as type guard 👍

ksaaskil: And in tests we can use any syntax our test environment uses, I think.

packages/unmock-core/src/__tests__/generator.test.ts
idantene: True, added originally by me.
The original intent was to be able to debug the different results (i.e. `expected "foo" received "14351"` vs `expected "true" received "false"`). It's a bit of a leftover at this point, would be nice to clean it up.

packages/unmock-core/src/__tests__/generator.test.ts
ksaaskil: Looking at the changes required in this test, I'm afraid it'll be quite painful to use `resp.body` for anything in tests if it's typed as `string | object`... I wonder if it was more user-friendly to do the thing you're not supposed to do and use `any` here for the `body`... 🤔

PR #191, opened 2019-08-26
add reset method

dangquangdon:
add to `src/index.ts` reset method, @ksaaskil , @mikesol , @idantene, please review


ksaaskil:
Note that you also need to add this to the interface exposed to the user, not only the concrete class.

idantene:
> Note that you also need to add this to the interface exposed to the user, not only the concrete class.

True! Good call.
In this case though (following #188), this might not be necessary. Good practice anyway :+1:

Issue #185, opened 2019-08-23
Property `method` in `UnmockRequest` should be enumerated


ksaaskil:
Property `method` is typed as `string` in `UnmockRequest` (alias for `ISerializedRequest`): 
```
export interface ISerializedRequest {
  ...
  method: string;
  ...
}
```
This is a bit annoying when making assertions on the request object, as one needs to guess if the correct method is  "get", "GET" or "Get". Change the typing to e.g. `HTTPMethod` from `service/interfaces.ts` (and add as export).

PR #194, opened 2019-08-28
Enumerate method in UnmockRequest

ksaaskil:
- Change `method` to be enumeration
- Move `HTTPMethod` and some other helpers one level higher as they're not service-specific


ksaaskil:
Resolves https://github.com/unmock/unmock-js/issues/185

idantene:
A bit of a late reply, looks good :+1: :+1:

PR #195, opened 2019-08-28
Decorate spy via codegen.

ksaaskil:
- Add helper methods such as `postRequestBody`, `getResponseBody`, etc. to the spy of type `ServiceSpy`
- Uses code generation to generate the methods, executed with `npm run generate-spy`.The code generation is not hooked up to any build step, so we'll have to manually keep the generation code and the generated code in sync. I think that's fine for now, later we could add a step in  CircleCI that compares the generated code to file contents and complains if there's a diff.
- Replaces #193


packages/unmock-core/src/service/spy/types.ts
```diff
@@ -0,0 +1,24 @@
+import { SinonSpy as SinonSpyType } from "sinon";
+import { ISerializedRequest, ISerializedResponse } from "../../interfaces";
+import { ISpyDecoration } from "./generated";
+
+export type SinonSpy = SinonSpyType<[ISerializedRequest], ISerializedResponse>;
+
+export interface ServiceSpy extends SinonSpy, ISpyDecoration {} // tslint:disable-line:interface-name
```
ksaaskil: Using an interface instead of `type ServiceSpy = ...` seems more user-friendly as interfaces can be extended.

ksaaskil: Well I guess nowadays one can extend type aliases as well, using an interface is only better here if someone wants to do declaration merging 🤔

PR #198, opened 2019-08-29
Include core in unmock package.

ksaaskil:
- So users can do `import unmock from "unmock"`.


packages/unmock/src/index.ts
```diff
@@ -0,0 +1,4 @@
+import unmock from "unmock-core";
+
+export * from "unmock-core";
+export default unmock;
```
ksaaskil: Maybe plain
```
export * from "unmock-core";
```
might suffice? I'd rather make the default explicit as I've seen it behave in weird ways 🤔

idantene: I second that decision. Prefer verbosity over implicitness (<- that does not feel like a proper word in English)

Issue #186, opened 2019-08-23
@types/sinon and other dependant types should be included as dependencies


ksaaskil:
Otherwise installers of `unmock` need to install those types themselves. Shipping types as "production" dependencies is the recommended approach [here](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies).

ksaaskil:
Dependencies to lerna packages can be added most easily with the [lerna add](https://github.com/lerna/lerna/tree/master/commands/add) command.

PR #201, opened 2019-08-30
Add types as dependencies.

ksaaskil:
- ~~Fixes the issue where `RequestResponseSpy` is typed as `any` because types from `sinon` are not included in the npm package~~ *EDIT: sinon types were added as dependencies in #203 so this can be a separate PR*
- See the discussion in #186


ksaaskil:
But if they're in the lerna level, then individual packages do not include their own type dependencies I assume? Or they include the type dependencies of all packages? Neither of those sounds right to me 🤔

idantene:
Fair points. Maybe it's just the dev dependencies then.

ksaaskil:
Waiting for the final approval and fixing looots of conflicts, probably easier to just start over as so many dependencies have come and gone... Why so, is this causing trouble?

ksaaskil:
So much has changed since this was implemented that it's easier to just start over.

PR #206, opened 2019-09-10
Unmock jest reporter.

ksaaskil:
- Add skeleton for `unmock-jest`
- Once published, can be installed and configured in Jest configuration with something like: `{ reporters: [["unmock-jest/reporter", { outputDirectory: "./reports" }]] }`
- Does not do anything yet other than log stuff to console
- Many libraries seem to add support for both test results processor and test reporter with small hacks (see e.g. [jest-junit](https://github.com/jest-community/jest-junit/blob/master/index.js#L63)), it seems to me that supporting the latter would be fine for now

TODO Later on
- [ ]  Add actual functionality
- [ ]  Add config validation with jest-validate
- [ ]  Add support for reading config from the environment
- [ ]  Decide on how the package depends on jest, just added it randomly as peer dependency with `>= 22.0.0` as [dts-jest](https://github.com/ikatyang/dts-jest/blob/master/package.json)


jest.config.base.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   testPathIgnorePatterns: ["/node_modules/", "/dist/"],
   moduleFileExtensions: [...defaults.moduleFileExtensions, "ts", "tsx"],
   moduleNameMapper: {
-    "^(unmock|unmock-cli|unmock-core)(?:/dist)?((?:/.*)|$)":
+    "^(unmock|unmock-cli|unmock-core|unmock-jest)(?:/dist)?((?:/.*)|$)":
```
idantene: No one:
My regex mind: We can stop editing this with `^(unmock(-[A-Za-z]+)?)(?:/dist)?((?:/.*)|$)`

mikesol: If it passes the build, we can assume the Lerna gods are appeased.

PR #207, opened 2019-09-10
Adds initial nock-like syntax

idantene:
This is done mostly as a building block for future PRs.
- Adds `unmock.nock(baseUrl, [name]).[HTTP METHOD](endpoint).reply([statusCode], [object/string])` allowing to dynamically add/update services.
- Many many `TODO` notes on where to improve, next steps, etc

### DIFF
--- a/packages/unmock-core/src/backend/index.ts
+++ b/packages/unmock-core/src/backend/index.ts
@@ -20,6 +20,7 @@ import {
 } from "../interfaces";
 import FSLogger from "../loggers/filesystem-logger";
 import { serializeRequest } from "../serialize";
+import { IObjectToService, ServiceStore } from "../service/serviceStore";
 import { resolveUnmockDirectories } from "../utils";
 import ClientRequestTracker from "./client-request-tracker";
 
@@ -92,7 +93,9 @@ interface IBypassableSocket extends net.Socket {
 }
 
 export default class NodeBackend {
-  private serviceStore: ServiceStoreType = {};
+  private static dummyStore = { services: {} };
+
+  private serviceStore: ServiceStore | undefined;
   private readonly config: INodeBackendOptions;
   private mitm: any;
 
@@ -101,7 +104,13 @@ export default class NodeBackend {
   }
 
   public get services(): ServiceStoreType {
-    return this.serviceStore;
+    return (this.serviceStore || NodeBackend.dummyStore).services;
+  }
+
+  public updateServices(newService: IObjectToService) {
+    if (this.serviceStore) {
+      this.serviceStore.updateOrAdd(newService);
+    }
   }
 
   /**
@@ -137,7 +146,7 @@ export default class NodeBackend {
       unmockDirectories,
     });
 
-    const { services, createResponse } = responseCreatorFactory({
+    const { serviceStore, createResponse } = responseCreatorFactory({
       listeners: [new FSLogger({ directory: this.config.servicesDirectory })],
       options,
       serviceDefLoader,
@@ -146,7 +155,7 @@ export default class NodeBackend {
     this.mitm.on("request", (req: IncomingMessage, res: ServerResponse) =>
       this.mitmOnRequest(createResponse, req, res),
     );
-    this.serviceStore = services;
+    this.serviceStore = serviceStore;
   }
 
   public reset() {
@@ -154,8 +163,13 @@ export default class NodeBackend {
       this.mitm.disable();
       this.mitm = undefined;
     }
-    Object.values(this.serviceStore).forEach(service => service.state.reset());
-    this.serviceStore = {};
+    if (this.serviceStore) {
+      // TODO - this is quite ugly :shrug:
+      Object.values(this.serviceStore.services).forEach(service =>
+        service.state.reset(),
+      );
+      this.serviceStore = undefined;
+    }
     ClientRequestTracker.stop();
   }
 
--- a/packages/unmock-core/src/backend/index.ts
+++ b/packages/unmock-core/src/backend/index.ts
@@ -109,8 +109,9 @@ export default class NodeBackend {
 
   public updateServices(newService: IObjectToService) {
     if (this.serviceStore) {
-      this.serviceStore.updateOrAdd(newService);
+      return this.serviceStore.updateOrAdd(newService);
     }
+    return undefined;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/unmock-core/src/backend/index.ts
```diff
@@ -101,7 +104,14 @@ export default class NodeBackend {
...
   public get services(): ServiceStoreType {
-    return this.serviceStore;
+    return (this.serviceStore || NodeBackend.dummyStore).services;
+  }
+
+  public updateServices(newService: IObjectToService) {
+    if (this.serviceStore) {
+      return this.serviceStore.updateOrAdd(newService);
+    }
+    return undefined;
```
ksaaskil: Is it expected one is "updating services" when no service store is defined?

packages/unmock-core/src/backend/index.ts
```diff
@@ -101,7 +104,14 @@ export default class NodeBackend {
   }
 
   public get services(): ServiceStoreType {
-    return this.serviceStore;
+    return (this.serviceStore || NodeBackend.dummyStore).services;
```
ksaaskil: I don't really understand the use for dummy store here, why not return `this.serviceStore && this.serviceStore.services || {}`?

packages/unmock-core/src/backend/index.ts
```diff
@@ -146,16 +156,21 @@ export default class NodeBackend {
...
     if (this.mitm) {
       this.mitm.disable();
       this.mitm = undefined;
     }
-    Object.values(this.serviceStore).forEach(service => service.state.reset());
-    this.serviceStore = {};
+    if (this.serviceStore) {
+      // TODO - this is quite ugly :shrug:
+      Object.values(this.serviceStore.services).forEach(service =>
+        service.state.reset(),
```
ksaaskil: On an unrelated note, should this rather be doing `service.reset()`?

packages/unmock-core/src/backend/index.ts
```diff
@@ -92,7 +93,9 @@ interface IBypassableSocket extends net.Socket {
 }
 
 export default class NodeBackend {
-  private serviceStore: ServiceStoreType = {};
+  private static dummyStore = { services: {} };
+
+  private serviceStore: ServiceStore | undefined;
```
ksaaskil: Should this rather be always defined but possibly empty?

packages/unmock-core/src/service/serviceStore.ts
```diff
@@ -1,8 +1,83 @@
...
+          },
+        },
+      },
+    };
+    const newUrls = [{ url: baseUrl }];
+
+    const newPaths = defaultsDeep(newPath, baseSchema.paths);
+    const newServers = defaultsDeep(newUrls, baseSchema.servers);
+    const finalSchema = { ...baseSchema, paths: newPaths, servers: newServers };
+    const newServiceCore = new ServiceCore({
```
ksaaskil: I'm not sure all of this building of the new `ServiceCore` is the service store's responsibility, maybe move it somewhere else?

packages/unmock-core/src/backend/index.ts
```diff
@@ -101,7 +104,14 @@ export default class NodeBackend {
   }
 
   public get services(): ServiceStoreType {
-    return this.serviceStore;
+    return (this.serviceStore || NodeBackend.dummyStore).services;
+  }
+
+  public updateServices(newService: IObjectToService) {
```
ksaaskil: Couldn't this just be `updateService(newService: ServiceCore)` to keep the backend simple?

packages/unmock-core/src/nock.ts
```diff
@@ -0,0 +1,98 @@
...
+  name?: string;
+}) => {
+  const dynFn = (method: HTTPMethod, endpoint: string) => ({
+    statusCode,
+    data,
+  }: {
+    statusCode: number;
+    data: Schema;
+  }) =>
+    backend.updateServices({
```
ksaaskil: Would it be rather possible to just fetch and/or update or create the service here to avoid too much of the http rest stuff go through the backend? The backend in my opinion should be something rather abstract.

packages/unmock-core/src/service/serviceStore.ts
ksaaskil: Maybe somewhere like `ServiceCore.from...`?

packages/unmock-core/src/backend/index.ts
idantene: Not sure - probably for a future PR. What should we do in that case?

packages/unmock-core/src/backend/index.ts
idantene: Just a typing shortcut, since typescript complains that an empty object does not contain the `services` key.

packages/unmock-core/src/backend/index.ts
idantene: It should - it would mean refactoring the backend so that it creates its own service store (where now it's a part of the `responseCreatorFactory` function, and it's assigned later on).

packages/unmock-core/src/service/serviceStore.ts
idantene: :+1:

packages/unmock-core/src/backend/index.ts
idantene: It could -- but I'm not sure how it would keep the backend any simpler than it is right now. It's essentially replacing one type with another (`ServiceCore` is not used in the backend).

At any case, we could change the type but we'd have to make `serviceStore` either external or accessible from outside the backend, since a new `ServiceCore` is only created once we possibly retrieved an already existing one.

packages/unmock-core/src/nock.ts
idantene: See my comment above - it essentially means exposing the service store from the backend. We could go down that path, although I'm not sure how that would look.

packages/unmock-core/src/backend/index.ts
ksaaskil: Ok, looked just a bit weird to me have a instruction being passed through the interceptor. But I don't fully understand anymore what refers what so maybe it's fine.

packages/unmock-core/src/nock.ts
ksaaskil: The confusion in updating the packages like this is one reason why requiring a simple `unmock.register(nockService)` or `unmock.addService(...)` would have made the code a lot simpler if we could afford that.

packages/unmock-core/src/backend/index.ts
idantene: I'll do that in a different PR; at the moment there's only `resetState()`, and I'm guessing it should also reset the spy?

packages/unmock-core/src/backend/index.ts
idantene: Added as a TODO note then :) will be part of refactoring

PR #209, opened 2019-09-10
Implements a crude runner

mikesol:
Currently only prints one error when shit hits the fan.
No way in the current architecture to elegantly pass options to the runner that percolate down to `json-schema-faker` (check out my hack in `generate.ts`...). I'd appreciate recommendations on how to make this cleaner. Once Kimmo finishes the reporter, we can hopefully link this to that.


packages/unmock-core/src/runner/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const defaultRunnerOptions: IRunnerOptions = {
+  maxLoop: 20,
+};
+
+const errorHandler = (
+  allFailed: boolean,
+  errors: Error[],
+  intermediaryErrors: Array<(string | { message: string})>,
+  cb?: jest.DoneCallback) => {
```
ksaaskil: TODO: This references `jest` types but the Jest types are not includes as `dependencies` of `unmock-core` so the build is not "self-contained" (also see https://github.com/unmock/unmock-js/pull/201).

packages/unmock-core/src/__tests__/runner.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      petstore = unmock.services.petstore;
+    });
+    afterAll(() => unmock.off());
+
+    beforeEach(() => {
+      petstore.reset();
+    });
+
+    test("runner loop works", runner(async () => {
+      petstore.state(withCodes(200));
```
ksaaskil: This is not even using the runner, or?

packages/unmock-core/src/runner/index.ts
mikesol: Yeah, we probably want to do some tests with mocha and also look into peer dependencies.

packages/unmock-core/src/__tests__/runner.test.ts
mikesol: `test("runner loop works", runner(async () => { `

PR #211, opened 2019-09-11
Add noop snapshotter.

ksaaskil:
- Add dependency to `expect`
- Add `IListenerInput` interface
- Add `FSSnapshotter` class, can only be instantiated as a singleton as having multiple snapshotters with different output directories is not possible (unless they all define a different extension on expect)
- Snapshotter does not do anything yet other than output to debug logs
- `FSSnapshotter` instance is a bit ugly state machine with `update` and `extendExpect` methods but this was the best way I could figure out to make it configurable and testable.


ksaaskil:
That's also possible! In the first phase I would maybe just avoid polluting the git repo too much as we don't probably want users to add the snapshots in VC (until we have some use case for that). And if services are pulled from npm (in future) or defined only via nock syntax, we may not even have a services directory 🤔

idantene:
Very valid points. I was thinking it might just be a bit easier to relate the `req/res` pair to a service that way. Then again, perhaps most users would rather use built-in functions rather than travel down their directory hierarchy...

PR #218, opened 2019-09-13
Add snapshot timestamp, export ISnapshot, fix folder missing bug

ksaaskil:
- Exports `ISnapshot` at top-level from `unmock` so it's usable in reporter. Would be nice to have it inside namespace such as `unmock.utils.snapshots.ISnapshot` but we don't organize our code like that atm so would be a bit more work.
- Adds timestamp to snapshots so they're easier to handle
- Adds check on write if the output folder (such as `${snapshotFolder}/435aAx/`) exists, as it may have been deleted by `deleteSnapshots()` method.

### DIFF
--- a/packages/unmock-core/src/__tests__/snapshotter.test.ts
+++ b/packages/unmock-core/src/__tests__/snapshotter.test.ts
@@ -4,6 +4,7 @@ import {
   format,
   ISnapshot,
   parseSnapshot,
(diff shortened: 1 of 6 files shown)


ksaaskil:
Sounds good, if we actually plan to do that though I'd rather add a new issue instead of adding TODO notes to the code 🤔

PR #219, opened 2019-09-13
More fluent nock syntax

idantene:
- Adds the option to chain different endpoints when defining a service, much like nock does (i.e. `unmock.nock(...).get(...).reply(...).post(...).reply(...)` etc.
**New info**
- The dynamic service creation will now return a "dynamic service specifcation", which one can build continuously on top of (see tests).
- Acccess to the service should *always* be done via `unmock.services.X`
- Updated so that multiple URLs are concatenated and not overwritten.

**Old info**
- **I'm not sure this is the best way to approach the idea.** There are two downsides:
    - We no longer return a `Service` object, but instead return an object that gives access to the `state` and `spy` properties of the newly created / modified `Service`.
    - Such calls can get confusing, as this becomes possible:
        ```typescript
        unmock.nock("https://foo.com/")
          .get("/bar")
          .reply(200, { content: u.str() })
          .post("/baz")
          .reply(404, { msg: u.str() })
          .state
            .get("/bar", { content: "bar!"})
            .post("/baz", { msg: "no reason given" });
        ```

### DIFF
--- a/packages/unmock-core/src/nock.ts
+++ b/packages/unmock-core/src/nock.ts
@@ -151,12 +151,12 @@ export class DynamicServiceSpec {
   public reply(
     statusCode: number,
     data?: InputToPoet | InputToPoet[],
-  ): FluentDyamicService;
-  public reply(data: InputToPoet | InputToPoet[]): FluentDyamicService;
+  ): FluentDynamicService;
+  public reply(data: InputToPoet | InputToPoet[]): FluentDynamicService;
   public reply(
     maybeStatusCode: number | InputToPoet | InputToPoet[],
     maybeData?: InputToPoet | InputToPoet[],
-  ): FluentDyamicService {
+  ): FluentDynamicService {
     if (maybeData !== undefined) {
       this.data = JSONSchemify(maybeData) as Schema;
       this.statusCode = maybeStatusCode as number;
@@ -179,7 +179,7 @@ export class DynamicServiceSpec {
   }
 }
 
-type FluentDyamicService = {
+type FluentDynamicService = {
   [k in HTTPMethod]: (endpoint: string) => DynamicServiceSpec;
 } & { state: StateType; spy: ServiceSpy };
 
@@ -188,7 +188,7 @@ const buildFluentNock = (
   baseUrl: string,
   name?: string,
   service?: Service,
-): FluentDyamicService => {
+): FluentDynamicService => {
   const dynFn = (method: HTTPMethod, endpoint: string) => ({
     statusCode,
     data,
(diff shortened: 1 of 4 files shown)


packages/unmock-core/src/nock.ts
```diff
@@ -166,58 +170,75 @@ export class DynamicServiceSpec {
...
-export const nockify = ({
-  backend,
-  baseUrl,
-  name,
-}: {
-  backend: NodeBackend;
-  baseUrl: string;
-  name?: string;
-}) => {
+type FluentDyamicService = {
```
ksaaskil: `Dynamic`, `Dyadic` or `Dyamic`? 🙂

packages/unmock-core/src/nock.ts
```diff
@@ -166,58 +170,75 @@ export class DynamicServiceSpec {
...
+  [k in HTTPMethod]: (endpoint: string) => DynamicServiceSpec;
+} & { state: StateType; spy: ServiceSpy };
```
ksaaskil: Should this have all the methods of `Service` or `IService` (whichever we expose) instead of hard-coding the properties it should have?

packages/unmock-core/src/nock.ts
```diff
@@ -166,58 +170,75 @@ export class DynamicServiceSpec {
...
+          code,
+          baseUrl,
+          name,
+        ),
+    }),
+    {},
+  ) as FluentDyamicService;
+  if (service !== undefined) {
+    fluent.state = service.state;
+    fluent.spy = service.spy;
```
ksaaskil: If you add new properties to service, you also have to remember to add them here I assume?

packages/unmock-core/src/nock.ts
ksaaskil: Like does it have `reset`?

mikesol:
My recommendation would be to

- Yield a valid service on `.build()`.
- Throw if a service is being defined for a URL on which `.build()` has been called already.
- On the object that `.build()` yields, have a `.release()` so that the above error won't be thrown (ie people can redefine the service, for example on a test-by-test basis).

packages/unmock-core/src/nock.ts
idantene: Whoops, thanks :D

packages/unmock-core/src/nock.ts
idantene: Yup, re my conversation with Mike; will remove these anyway so it'll be more like
```
const dynamicServiceSpec = unmock.nock(....).get(...).reply(...).post(...).reply(...)
const service = unmock.services[...]
```
That will avoid the confusion.

PR #222, opened 2019-09-16
Uses simpler API for setting the state

mikesol:
Well, so much for not working on this. Couldn't help it - late night coding binge again :-D

I've put this behind a feature flag. It is for internal use only, but would be good to merge as `package-lock.json` had to update a lot of dependencies because there were `peerDependency` issues.

Migrating the state API to this should be fairly easy. I wrote `transformer.test.ts` such that it has feature parity with `state.test.ts`, so we can come up with functions that take the state DX (`$code`, `$times` etc) and translate it into the new system.

I prefer this way of dealing with the state, as it is:

- simple (only one API: `(req: ISerializedRequest, o: OpenAPI) => OpenAPI`
- composable (in the tests, you just need to use `reduce` to combine multiple transformers together)
- extensible (you can more or less do whatever you want)

I would recommend merging this and then changing `transformer` to `state`, getting rid of the `state` code, and using helper functions to accept the old `state` syntax. Once that is done, we can cut out about 16 tests as they will just be redundant.

While this adds `generator-experimental.ts` which is 350 lines of code, it will allow us to delete multiple files and over 1000 lines of code, so it is also a maintainability win IMO.


packages/unmock-core/src/generator-experimental.ts
```diff
@@ -0,0 +1,328 @@
...
+  Parameter,
+  PathItem,
+  Reference,
+  Response,
+  Responses,
+  Schema,
+} from "./service/interfaces";
+import { ServiceStore } from "./service/serviceStore";
+
+export const without = <T>(p: T, m: keyof T): T => {
```
ksaaskil: Same as [_.omit](https://lodash.com/docs/4.17.15#omit)?

packages/unmock-core/src/generator-experimental.ts
```diff
@@ -0,0 +1,328 @@
...
+const prunePathItemInternal = (m: MethodNames, a: MethodNames[], p: PathItem): PathItem =>
+  a.length === 0 ? p : prunePathItemInternal(m, a.slice(1), a[0] === m ? p : without(p, a[0]));
+
+export const prunePathItem = (m: MethodNames, p: PathItem) => prunePathItemInternal(m, allMethods, p);
+
+export const matchesInternal = (a: string[], b: string[]): boolean =>
+  a.length === b.length
+  && (a.length === 0
+    || (((a[0] === b[0])
+      || (b[0].length > 2 && b[0][0] === "{" && b[0].slice(-1) === "}")) && matchesInternal(a.slice(1), b.slice(1))));
```
ksaaskil: These internal methods could maybe use some documentation and/or examples, it's quite hard to follow what they're doing 🤔

packages/unmock-core/src/generator-experimental.ts
```diff
@@ -0,0 +1,328 @@
...
+    fromTraversable(array)<[string, OpenAPIObject]>()
+      .filter(([__, o]) =>
+        matchUrls(req.protocol, req.host, o).length > 0))
+  .composeLens(valueLens()).modify(oai => f({
+    ...req,
+    path: truncatePath(req.path, oai, req),
+    pathname: truncatePath(req.pathname, oai, req),
+  }, oai))(r);
+
+export function responseCreatorFactory2({
```
ksaaskil: This is a very long function, would it be possible to split this into smaller logical units?

packages/unmock-core/src/generator-experimental.ts
```diff
@@ -0,0 +1,328 @@
...
+          a => a,
+        ))
+        .composeIso(new Iso<[string, Header], Parameter>(
+          a => ({ ...a[1], name: a[0], in: "header"}),
+          a => [a.name, a],
+        ))
+        .composeOptional(parameterSchema(schema.value))
+        .composeGetter(identityGetter())
+        .getAll(operation.value);
+      const bodySchema = path1
```
ksaaskil: All these composed operations would in my opinion need some comments, they're quite abstract and hard to debug for anyone not familiar with monocle.

packages/unmock-core/src/generator.ts
```diff
@@ -32,6 +32,7 @@ function firstOrRandomOrUndefined<T>(arr: T[]): T | undefined {
     ? arr[0]
     : arr[Math.floor(Math.random() * arr.length)];
 }
+export const USE_EXPERIMENTAL_GENERATOR = { yes: false };
```
ksaaskil: Maybe allow turning it on by an environmental variable, it might be easier to test than with in-code booleans?

packages/unmock-core/src/service/interfaces.ts
```diff
@@ -100,6 +101,9 @@ export interface IObjectToService {
 }
 
 export interface IServiceCore {
+
+  transformer: (req: ISerializedRequest, o: OpenAPIObject) => OpenAPIObject;
```
ksaaskil: Is this rather a `transform` operation? Also wondering if this should be curried or not, without currying it's definitely easier to understand but just wondering if it would be more convenient.

package.json
```diff
@@ -53,6 +55,7 @@
     "@types/xregexp": "^3.0.30",
     "axios": "^0.19.0",
     "codecov": "^3.3.0",
+    "deep-equal": "^1.1.0",
```
ksaaskil: Is this somehow better than lodash `_.isEqual`?

packages/unmock-core/package.json
```diff
@@ -43,5 +46,9 @@
     "uuid": "^3.3.2",
     "winston": "^3.2.1",
     "xregexp": "^4.2.4"
+  },
+  "devDependencies": {
+    "@types/deep-equal": "^1.0.1",
+    "deep-equal": "^1.1.0"
```
ksaaskil: We've kept each package's devDependencies at the top-level package.json, I don't see any reason to put them here (and it does not work as expected I think).

packages/unmock-core/src/generator-experimental.ts
mikesol: Ouch, yeah, it is quite long. Hm, difficult to know what the logical units would be, as everything kind of builds on everything else...

One thing that is possible is making all of those lenses functions that create a lens. That'd reduce the size by around 60-70 lines, but perhaps at the expense of readability. Not sure what the best way forward is.

packages/unmock-core/src/generator-experimental.ts
mikesol: Hm, yeah that is definitely a danger. The problem is that this is tightly coupled with `monocle-ts`. I'm not sure how to fix it, as it would essentially be documenting monocle. Maybe a comment on top of the file introducing people to monocle so that a maintainer felt comfortable with the main abstractions (Optional, Getter, Lens, Prism, Traversal, Iso).

packages/unmock-core/src/service/interfaces.ts
mikesol: I had the exact same internal debate and arrived at the exact same question. I don't have an opinion...

packages/unmock-core/src/generator.ts
mikesol: I'll leave it as this for now, just because otherwise if we flipped the switch and used `state` it may wreak havoc. As the flag will be taken out soon-ish, I'm comfortable keeping it this way.

packages/unmock-core/src/generator-experimental.ts
ksaaskil: I'm not saying one should explain what monocle does, but I'm not sure even someone familiar with `monocle-ts` would see what this doing...:
```ts
      const headerSchema1 = path1
        .composeOptional(Optional.fromNullableProp<Response>()("headers"))
        .composeIso(objectToArray())
        .composeTraversal(fromTraversable(array)())
        .composePrism(new Prism<[string, Reference | Header], [string, Header]>(
          a => isReference(a[1])
            ? stringHeader(a[0], getHeaderFromRef(schema.value, refName(a[1])))
            : some([a[0], a[1]]),
          a => a,
        ))
        .composeIso(new Iso<[string, Header], Parameter>(
          a => ({ ...a[1], name: a[0], in: "header"}),
          a => [a.name, a],
        ))
        .composeOptional(parameterSchema(schema.value))
        .composeGetter(identityGetter())
        .getAll(operation.value);
```

ksaaskil: And while I agree with the notion that comments can be evil and the code should document itself, I think some help would maybe be in-place here and there 🙂

mikesol: Yeah, it's a really good point. The way I do it is hovering my mouse over in the IDE, seeing what comes in and out, and looking at the types of the schema to see how the optic is drilling down. What would help me is a suggestion for how to do the comments. I completely agree that they are necessary: if I saw this without an IDE, I would not touch this code, and even in an IDE where I get the types, I would be really intimidated.

ksaaskil: It's not worth commenting on every line, but maybe add a simple comment in every assignment (like they were isolated functions of their own) shortly explaining what it's doing and what's the example result, something like
```ts
/**
Extract header from...
Example result: ...
*/
const headerSchema1 = path1...
```
Also using more descriptive names instead of `a` in anonymous functions might help a bit. All this coming from someone with no experience in monocle-ts but just interested in reading the code from the perspective of unmock-js functionality.

mikesol: Changed, great idea!

PR #228, opened 2019-09-16
Changes state to new state

mikesol:
Explores impact on code base of changing state to the new method.
The biggest global change I see is that a lot of previous failures are now no-ops, meaning that setting the state to something bogus fails silently. This can be fixed using the strategy in `state.test.ts` of throwing if the input is equal to the output. It can also be fixed by using `modifyOption` instead of `modify` in `openapi-refinements` and throwing on `none`.


packages/unmock-core/src/__tests__/backend/index.test.ts
```diff
@@ -1,4 +1,5 @@
 import axios from "axios";
+import { removeCodes } from "openapi-refinements";
```
ksaaskil: `openapi-refinements` is such an integral part of our whole state management that I maybe would have added it as a Lerna repo in `unmock` for easier maintenance and development (until it's stable enough).

packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -1,140 +1,164 @@
...
+import { ISerializedRequest } from "../interfaces";
 
-const mockOptions = {
-  flaky: () => false,
-  isWhitelisted: (_: string) => false,
-  log: (_: string) => undefined,
-  useInProduction: () => false,
-};
+test("hasUrl works", () => {
+  expect(matchUrls(
```
ksaaskil: I think it would be nice if each of the expects was a separate test under a common `describe()`, makes it easier to see what is actually being tested in each case. But it's not important now, just a general comment.

packages/unmock-core/src/__tests__/generator.test.ts
```diff
@@ -1,140 +1,164 @@
...
-      protocol: "http",
-      query: {},
-    });
-    expect(resp).toBeDefined();
-    if (resp && resp.body !== undefined) {
-      JSON.parse(resp.body).forEach((pet: any) => expect(pet.id).toEqual(1));
-    } else {
-      throw new Error("Response body was undefined?");
-    }
+test("hoist transformer works", () => {
```
ksaaskil: So many of the "stuff works" tests here 😬  I think it would also act as documentation if the tests were more like
```js
describe("hoist transformer", () => {
  it("should hoist correctly when matching host", () => { ... });
  it("should result in no-op when not matching host, () => { ... });
});
```

packages/unmock-core/src/generator-utils.ts
```diff
@@ -0,0 +1,244 @@
...
+  oneOfReject,
+  pathParameter,
+  removeCodes,
+  requestBody,
+  responseBody,
+} from "openapi-refinements";
+export {
+  Arr,
+  Addl,
+} from "openapi-refinements";
```
ksaaskil: Reaaally strong coupling to `openapi-refinements` and we don't yet know if that's useful for people as a separate library, including this in lerna would make development and maintenance faster in my opinion.

packages/unmock-core/src/interfaces.ts
```diff
@@ -1,8 +1,11 @@
...
...| 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 419 | 420 | 421 | 422 | 423 | 424 | 425 | 426 | 427 | 428 | 429 | 430 | 431 | 432 | 433 | 434 | 435 | 436 | 437 | 438 | 439 | 440 | 441 | 442 | 443 | 444 | 445 | 446 | 447 | 448 | 449 | 450 | 451 | 452 | 453 | 454 | 455 | 456 | 457 | 458 | 459 | 460 | 461 | 462 | 463 | 464 | 465 | 466 | 467 | 468 | 469 | 470 | 471 | 472 | 473 | 474 | 475 | 476 | 477 | 478 | 479 | 480 | 481 | 482 | 483 | 484 | 485 | 486 | 487 | 488 | 489 | 490 | 491 | 492 | 493 | 494 | 495 | 496 | 497 | 498 | 499 | 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 509 | 510 | 511 | 512 | 513 | 514 | 515 | 516 | 517 | 518 | 519 | 520 | 521 | 522 | 523 | 524 | 525 | 526 | 527 | 528 | 529 | 530 | 531 | 532 | 533 | 534 | 535 | 536 | 537 | 538 | 539 | 540 | 541 | 542 | 543 | 544 | 545 | 546 | 547 | 548 | 549 | 550 | 551 | 552 | 553 | 554 | 555 | 556 | 557 | 558 | 559 | 560 | 561 | 562 | 563 | 564 | 565 | 566 | 567 | 568 | 569 | 570 | 571 | 572 | 573 | 574 | 575 | 576 | 577 | 578 | 579 | 580 | 581 | 582 | 583 | 584 | 585 | 586 | 587 | 588 | 589 | 590 | 591 | 592 | 593 | 594 | 595 | 596 | 597 | 598 | 599;
```
ksaaskil: Could this be created from a list like the rest methods are created, like
```
const RESTMethodTypes = [
  "get",
  "head",
  "post",
  "put",
  "patch",
  "delete",
  "options",
  "trace",
] as const;
export type HTTPMethod = typeof RESTMethodTypes[number];
```
?

ksaaskil:
And maybe please squash merge when merging, makes it easier to keep track of changes at higher-level. Not sure if there are any individual commits here that would be useful to refer to.

PR #237, opened 2019-09-18
Prettier

mikesol:
Uses prettier instead of tslint as per Carolyn's recommendation


ksaaskil:
As discussed previously, `prettier` and `tslint` serve very different purposes? Prettier formats code, linter complains about code style like unused variables. So I think tslint should rather be replaced with [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint).

ksaaskil:
```
Code-quality rules: eg no-unused-vars, no-extra-bind, no-implicit-globals, prefer-promise-reject-errors...

Prettier does nothing to help with those kind of rules. They are also the most important ones provided by linters as they are likely to catch real bugs with your code!
```
https://prettier.io/docs/en/comparison.html

ksaaskil:
There's an example configuration for eslint in [unmock.github.io](https://github.com/unmock/unmock.github.io).

mikesol:
Closed as it goes too far - we need a linter as well...

carolstran:
`I think prettier and tslint serve very different purposes?` Yes! I never said we needed to replace tslint with Prettier, I just don't like the company behind tslint and mentioned that it'd be cool if we could find an alternative 😂 

That being said, having a code formatter like Prettier would be great if not already implemented as it makes managing outside contributions easier buuuuut we can discuss this more once I actually join 👐

PR #246, opened 2019-09-19
Add more request-response accessors.

ksaaskil:
- Simplified tests
- [ ]  Generator contains a lot of duplicate code, use a helper function to remain dry
- The names `IIncomingHeaders` and `IOutgoingHeaders` are bit unclear, maybe better rename them as `IRequestHeaders` and `IResponseHeaders`?


packages/unmock-core/src/service/spy/generated/index.ts
```diff
@@ -3,24 +3,65 @@
  */
 import { match, SinonMatcher } from "sinon";
```
mikesol: Remind me why we're checking this into VC? I'm sure there's a good reason, but I forget what the decision was.

ksaaskil: I think it's just nicer to manage, changing it is a conscious operation instead of requiring repo pullers to generate that themselves as part of the build.

PR #249, opened 2019-09-19
Support consecutive calls to `reply` in nock syntax

idantene:
- Adds support for consecutive calls to `reply` in nock syntax
- Requires us to update interfaces etc once we add e.g. `replyWithFile`, or headers, etc
- Simplifies some logic
- Adds test and fixes some old ones
- Adds comments
- Has a very verbose branch name

### DIFF
--- a/packages/unmock-core/src/__tests__/backend/state.test.ts
+++ b/packages/unmock-core/src/__tests__/backend/state.test.ts
@@ -1,14 +1,15 @@
 import axios from "axios";
+import { OpenAPIObject } from "loas3/dist/generated/full";
+import { Arr } from "openapi-refinements";
 import * as path from "path";
 import { Service, UnmockPackage } from "../..";
 import NodeBackend from "../../backend";
-import { Arr } from "openapi-refinements";
-import { OpenAPIObject } from "loas3/dist/generated/full";
 import { transform } from "../../generator-utils";
 import { ISerializedRequest } from "../../interfaces";
 
 const {
   withCodes,
+  withoutCodes,
   mapDefaultTo,
   responseBody,
   noopThrows,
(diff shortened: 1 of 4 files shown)


packages/unmock-core/src/__tests__/backend/state.test.ts
```diff
@@ -82,37 +83,35 @@ describe("Node.js interceptor", () => {
...
+        withoutCodes("default"),
+        responseBody({ lens: ["message"] }).const("Hello World"),
       );
       const response = await axios("http://petstore.swagger.io/v1/pets");
       expect(response.status).toBe(200);
       expect(response.data.message).toEqual("Hello World");
     });
 
-    test("gets correct state after multiple overriden state requests", async () => {
+    test.skip("gets correct state after multiple overriden state requests", async () => {
```
idantene: Skipped as this doesn't work yet, but it outlines what we'd like to achieve

packages/unmock-core/src/__tests__/backend/state.test.ts
```diff
@@ -152,15 +151,6 @@ describe("Node.js interceptor", () => {
       }
     });
 
-    test("sets an entire response from function", async () => {
```
idantene: Embedded in the next test (`sets an entire response from with request object`)

packages/unmock-core/src/__tests__/backend/state.test.ts
```diff
@@ -174,40 +164,6 @@ describe("Node.js interceptor", () => {
       expect(response.data).toBe("petstore.swagger.io");
     });
 
-    test("sets an entire response from function with DSL", async () => {
```
idantene: Has no meaning with the new state management.

packages/unmock-core/src/__tests__/dynamicService.test.ts
```diff
@@ -88,5 +88,25 @@ describe("Tests dynamic path tests", () => {
       .nock("https://abc.com", "foo")
       .get("foo")
       .reply(500);
+    // @ts-ignore // "private fields" heh...
+    expect(unmock.services.foo.core.oasSchema.servers).toEqual([
```
ksaaskil: Maybe services could expose which servers they apply to so one would not need to go this deep into internals.

packages/unmock-core/src/nock.ts
```diff
@@ -198,74 +188,103 @@ type UpdateCallback = ({
 type Primitives = string | number | boolean;
 type InputToPoet = { [k: string]: any } | Primitives | Primitives[];
 
-export class DynamicServiceSpec {
+// How the fluent dynamic service API looks like (e.g. specifies `get(endpoint: string) => DynamicServiceSpec`)
+type FluentDynamicService = {
+  [k in HTTPMethod]: (endpoint: string) => DynamicServiceSpec;
+};
+
+// How the actual dynamic service spec looks like (e.g. `reply(statusCode: number, data: InputToPoet): ...`)
```
ksaaskil: Not sure how necessary these comments are 🙂

packages/unmock-core/src/nock.ts
```diff
@@ -198,74 +188,103 @@ type UpdateCallback = ({
...
   public reply(
-    maybeStatusCode: number | InputToPoet | InputToPoet[],
+    maybeStatusCode: CodeAsInt | "default" | InputToPoet | InputToPoet[],
     maybeData?: InputToPoet | InputToPoet[],
-  ): FluentDynamicService {
+  ): FluentDynamicService & IDynamicServiceSpec {
     if (maybeData !== undefined) {
-      this.data = JSONSchemify(maybeData) as Schema;
-      this.statusCode = maybeStatusCode as number;
+      this.data = JSONSchemify(maybeData) as Schema; // TODO should this be some JSSTX?
```
ksaaskil: Is the code a good place for TODOs like this, are we planning to change to JSSTX at some point? 🤔

packages/unmock-core/src/__tests__/dynamicService.test.ts
idantene: I'd rather keep those things abstracted away from the end-user.
This was just the simplest way to test it in *this* file (the other way would be to put this in a backend test and then just try and access both the URLs).

packages/unmock-core/src/nock.ts
idantene: Good question. `JSSTX` elements are all over the place (Mike's own `JSSTAnything, JSSTEmpty`, etc). I'm a bit tempted for us to get rid of OAS at this stage, and `Schema` comes from `loas`. Just wanted to put it there for later review.

ksaaskil: Ok, I'd maybe just keep those in issues then instead of adding bunch of TODOs in the code. But it doesn't matter.

PR #250, opened 2019-09-19
Randomize optionals and minItems + clean up

idantene:
- Use the `s/seedHack/runnerHack/g` to include more options for randomization and generation.
- Specifically, increase `minItems` by a randomized, exponential increase, and add `optionalsProbability` (randomized at each run).
- Remove old serializers and deserializers which weren't in use anyway.
- Remove `service/constants.ts`, `service/constants.ts` and `backend/constants.ts` which weren't in use anyway.
- Remove `XRegExp` from dependency (should we have a sweep over all dependencies in general?)


packages/unmock-core/src/generator.ts
```diff
@@ -52,8 +52,15 @@ import {
   Schema,
 } from "./service/interfaces";
 import { ServiceStore } from "./service/serviceStore";
-export const seedHack = {
+export const runnerHack = {
```
ksaaskil: I would maybe just name these as "runnerConfiguration" or something more sensible 🙂

packages/unmock-core/src/generator.ts
```diff
@@ -52,8 +52,15 @@ import {
...
   seed: 0,
+  optionalsProbability: 1.0,
+  minItems: 0,
+};
+export const resetRunnerHack = () => {
```
ksaaskil: Something like `runner.configuration = DEFAULT_CONFIGURATION` would maybe reduce the need to remember to define these in two places.

idantene: It's actually meant to be used within the generator, to reset `minItems`.
I can't decide if it's better after generation or after `runner` is complete. But TBH I don't understand yet how `runner` works, so I preferred putting it here :)
(see line num 727)

idantene: Now that I do, I'll just move it there.

PR #259, opened 2019-09-20
Use CSS for switching between test suites, rewrite components with React.

ksaaskil:
- One can now switch between test suites by clicking on the labels, implemented with CSS radio-buttons and sibling selectors
- Rewrote the test suite components with React, xmlbuilder can soon be removed
- [ ]  Would be nice to add checkbox CSS hacks also to individual tests so one can enlarge and collapse them
- [ ]  Needs tests
- [ ]  Full paths to test files are a bit annoying, could drop longest common directory

Looks like this:
Screenshot 2019-09-20 at 11 05 03

### DIFF
--- a/packages/unmock-jest/src/reporter/components/test-suite.tsx
+++ b/packages/unmock-jest/src/reporter/components/test-suite.tsx
@@ -1,5 +1,7 @@
+import { map } from "lodash";
 import * as React from "react";
 import { ITestSuite } from "../types";
+import Test from "./test";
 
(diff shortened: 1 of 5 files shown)


ksaaskil:
I can remove some of the returns when cleaning up, I'm used to always writing new React components with returns so it's quick to add logic and have the return be an explicit "render" function.

ksaaskil:
Nothing really, I'm just using `mapValues` here and there and `map` is a remnant of that, can remove.

PR #263, opened 2019-09-20
Accept an array as path for nock-syntax, support regex (named and unnamed).

### LINKED ISSUE #252: Allow the path specification with an array
As per https://github.com/unmock/unmock-js/tree/new-readme#specifying-path.
Also, the more I think about it, the more the regex in the array should be a key-value pair so that the `responseBody` override should be able to reference it. So...

```typescript
["id",["foo", /[123]+/],"bar"]
```

Instead of 

```typescript
["id", /[123]+/, "bar"]
```

The former allows for `responseBody({ path: "/id/{foo}"})` whereas the latter does not. Of course, we should support the latter as well, in which case we just generate a random name for the path parameter (ie a uuid).

idantene:
Resolves #252.
Was also going over `generator.ts` and to get hands-on experience I changed small stuff there.

Todo for future:
- [ ] Support wildcards?
- [ ] Update state to include specialization for paths with parameters, where the parameter is noted. Similarly, support for wildcards (`?`, `*`, `**`).

### DIFF
--- a/packages/unmock-core/src/service/serviceCore.ts
+++ b/packages/unmock-core/src/service/serviceCore.ts
@@ -32,25 +32,25 @@ export class ServiceCore implements IServiceCore {
     // Create a new endpoint from array if needed
     const endpointParameters = {
       parameters: Array.isArray(endpoint)
-        ? endpoint
-            .filter(e => typeof e !== "string")
-            .map(e => ({
-              in: "path",
-              required: true,
-              ...(e instanceof RegExp
-                ? {
-                    name: Math.random()
-                      .toString(36)
-                      .substring(2),
-                    schema: { pattern: e.toString() },
-                  }
-                : {
-                    name: e[0],
-                    schema: {
-                      pattern: e[1].toString(),
-                    },
-                  }),
-            }))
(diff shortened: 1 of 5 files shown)


packages/unmock-core/src/service/serviceCore.ts
```diff
@@ -29,10 +29,59 @@ export class ServiceCore implements IServiceCore {
     const mediaType =
       typeof response === "string" ? "text/*" : "application/json";
     // TODO: Decouple from ServiceCore :( - this is nasty
+    // Create a new endpoint from array if needed
+    const endpointParameters = {
```
ksaaskil: Any examples available for what this does? 😁

packages/unmock-core/src/service/serviceCore.ts
```diff
@@ -29,10 +29,59 @@ export class ServiceCore implements IServiceCore {
...
+              : {
+                  name: e[0],
+                  schema: {
+                    pattern: e[1],
+                  },
+                }),
+          }))
+        : [],
+    };
+    const newEndpoint = Array.isArray(endpoint)
```
ksaaskil: Also this is a bit hard to read to see what they do but looks good!

PR #269, opened 2019-09-22
Makes after to counteract times

mikesol:
This way, a state manipulation will be able to start after a certain number of times.


packages/unmock-core/src/generator-utils.ts
```diff
@@ -280,6 +280,19 @@ export const transform = {
       return out;
     };
   },
+  after: (n: number) => (f: IStateTransformer) => {
+    const counter = { c: 0 };
```
ksaaskil: Is this somehow better than `let counter = 0`? 🙂

PR #271, opened 2019-09-24
Improve tldr method

idantene:
- Assign "correct" status codes for HTTP methods on `tldr` (instead of 200 for everything)
- Assign all HTTP methods for `tldr` (adds `head`, `options`, `trace`, etc)
- Simplify/shorten the calls to add more complicated paths

### DIFF
--- a/packages/unmock-core/src/nock.ts
+++ b/packages/unmock-core/src/nock.ts
@@ -215,15 +215,15 @@ type InputToPoet = { [k: string]: any } | Primitives | Primitives[];
 
 // How the fluent dynamic service API looks like (e.g. specifies `get(endpoint: string) => DynamicServiceSpec`)
 export interface IFluentDynamicService {
-  tldr: () => void;
-  get: (endpoint: ValidEndpointType) => DynamicServiceSpec;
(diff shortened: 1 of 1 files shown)


packages/unmock-core/src/nock.ts
```diff
@@ -215,15 +215,15 @@ type InputToPoet = { [k: string]: any } | Primitives | Primitives[];
...
-  tldr: () => void;
-  get: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  head: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  post: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  put: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  patch: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  delete: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  options: (endpoint: ValidEndpointType) => DynamicServiceSpec;
-  trace: (endpoint: ValidEndpointType) => DynamicServiceSpec;
+  tldr(): void;
```
idantene: Better described without arrow notation, as this way these are declared as methods; with arrow notation, they're declared as properties/attributes.

PR #274, opened 2019-09-25
Jsf bug

mikesol:


packages/unmock-core/src/runner/index.ts
```diff
@@ -63,8 +62,16 @@ export default (
...
       const r = await (fn ? fn(intermediaryDoneCallback) : undefined);
+      // reset histories
+      Object.entries(backend.serviceStore.services).forEach(([_, service]) => {
+        service.spy.resetHistory();
+      });
       res.push(r);
     } catch (e) {
+      // reset histories
+      Object.entries(backend.serviceStore.services).forEach(([_, service]) => {
+        service.spy.resetHistory();
```
ksaaskil: Would be maybe better to just add a method `serviceStore.resetSpies()` 🙂

packages/unmock-core/src/runner/index.ts
```diff
@@ -9,16 +9,15 @@ const defaultRunnerOptions: IRunnerOptions = {
...
   intermediaryErrors: Array<string | { message: string }>,
   cb?: jest.DoneCallback,
 ) => {
   // tslint:disable-next-line:max-line-length
   const msg = (rest: string) =>
-    `${chalk.red("@unmock")} - ${
-      allFailed ? "all" : "some"
-    } tests failed. Here's the first error.\n${rest}`;
+    `This many tests failed: ${errors.length +
+      intermediaryErrors.length} out of ${nTimes}. Here's the first error.\n${rest}`;
```
ksaaskil: Some reason to remove `chalk` from here?

packages/unmock-core/src/runner/index.ts
ksaaskil: Also, could use a `finally` block?

packages/unmock-core/src/runner/index.ts
mikesol: Did not make sense to keep it in as I'm not sure what to chalkify. If there is something that deserves to leap out, of course we should add it back.

packages/unmock-core/src/runner/index.ts
mikesol: I added a finally block. We can add `serviceStore.resetSpies()` in another commit.
In general, @idantene was right that the code has big technical debt on the topmost layer because of the whole `backend` zombie alive from the `jsdom` days. If a generous soul wants to refactor during Hacktober fest so that the code is more readable and maintanable given its intent, that would be awesome!

ksaaskil: Yes, I also think it should have been renamed to `Interceptor` in those days when it was only a MITM wrapper to ensure it only does interception and not so much more.

PR #275, opened 2019-09-25
Version 0.3.5

mikesol:


mikesol:
> Would be nice if every release had links to the corresponding changes in docs and examples but that's the ideal world 🙂

Yeah, agreed - it'd be even nicer if there were a build step that upgraded "official" documentation and made sure that there was no breakage.

PR #289, opened 2019-09-27
Use Faker in jsf

idantene:
Depending on https://github.com/unmock/json-schema-poet/pull/7
- [x] Update poet version


idantene:
Sure, no harm in adding a sanity test.

PR #293, opened 2019-09-28
Adds scaffolding for redactor

mikesol:
This is a basic sketch for redactable test reports. Not sure if the interface is the best, and not sure if we want redaction on the snapshot level as well, but IMO this achieves minimal viable redaction in the reporter, which was a common concern in Budapest.


ksaaskil:
It would be simpler if the redactor was just part of e.g. the snapshot loading phase or toTestSuites function, I don't see a reason to pass it around in the code...

mikesol:
👍 we can close the PR if there is a better way to do it. I don't really know that part of the code base well enough to know what the best practices are, so if there is something cleaner & simpler then feel free to make a PR with that. Also, I agree with you that we should have snapshot redaction. We should actually IMO have both if possible, ideally with the same interface

ksaaskil:
Replaced by #295 .

PR #294, opened 2019-09-28
Optional api

mikesol:
This adds the following three members to the `u` API:
- `u.opt`, which makes an element of an object optional
- `u.fuzz`, which takes an object and converts any constants to fuzzable values (ie `"bob"` becomes `u.string()`). in the `fuzz` context, constants are represented with `u.const()`
- `u.unfuzz`. this reverts the `fuzz` context.

### DIFF
--- a/packages/unmock-core/src/nock.ts
+++ b/packages/unmock-core/src/nock.ts
@@ -1,3 +1,5 @@
+import { array } from "fp-ts/lib/Array";
+import { none, Option, some } from "fp-ts/lib/Option";
 import * as io from "io-ts";
 import { cnst_, extendT, tuple_, type_ } from "json-schema-poet";
 import {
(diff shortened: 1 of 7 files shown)


packages/unmock-core/src/nock.ts
```diff
@@ -83,8 +114,59 @@ export type ExtendedValueType =
...
+            q: string,
+            r: ExtendedValueType | IMaybeJSONValue,
+          ): Option<[string, ExtendedValueType]> =>
+            MaybeJSONValue.is(r) ? none : some([q, r]))(a[0], a[1]),
+        a => a,
+      ),
+    )
+    .composeGetter(identityGetter())
+    .getAll(i)
+    .reduce((a, b) => ({ ...a, [b[0]]: b[1] }), {});
```
ksaaskil: Would `asFold()` + folding be more idiomatic `monocle-ts` than `getAll(i)` +  `reduce`? Not important, just a comment.

mikesol: I need to learn more about folds...

PR #298, opened 2019-10-01
Remove the header used internally from request.

ksaaskil:
- `UnmockRequest` contained the `x-unmock-req-id` header that is used internally fir mapping between `IncomingMessage` and `ClientRequest`, it should be removed when not used anymore


ksaaskil:
Yeah, seemed like too much effort compared to benefits in this case to clone the `IncomingMessage` object and return the copy.

PR #306, opened 2019-10-08
Split interceptor from backend.

ksaaskil:
- Create new `IInterceptor` interface. Creating an interceptor starts intercepting requests, so I also added `IInterceptorConstructor` to add typing to interceptor constructor
- First step in isolating Node.js-related things from logic not related to Node.js
- Also reduces the responsibilities of the now-big `NodeBackend` class. I think that could evolve to become simply simply the common `Backend` using `IInterceptor` and `IServiceDefLoader` created dynamically.


packages/unmock-core/src/backend/index.ts
```diff
@@ -140,67 +117,41 @@ export default class NodeBackend {
...
-    this.mitm = Mitm();
-
-    ClientRequestTracker.start();
 
-    // Client-side socket connect, use to bypass connections
-    this.mitm.on("connect", (socket: IBypassableSocket, opts: RequestOptions) =>
-      this.mitmOnConnect(options, socket, opts),
-    );
+    if (this.interceptor) {
+      this.interceptor.disable();
```
idantene: Where is this assigned, and/or why set it to `undefined`? Should the interceptor just have a no-op if we call `disable` repeatedly?

packages/unmock-core/src/interceptor/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export interface IInterceptor {
+  disable(): void;
+}
+
+/**
+ * Create an interceptor and start intercepting requests.
+ * @param options
+ */
+export function createInterceptor(options: IInterceptorOptions): IInterceptor {
```
idantene: Is this here for the purpose of being a future factory function, or encapsulating the `NodeInterceptorConstructor`?

packages/unmock-core/src/interceptor/node-interceptor.ts
```diff
@@ -0,0 +1,106 @@
...
+
+interface IBypassableSocket extends net.Socket {
+  bypass: () => void;
+}
+
+/**
+ * Node.js interceptor using node-mitm.
+ * @param options Interceptor options
+ */
+const NodeInterceptorConstructor: IInterceptorConstructor = class NodeInterceptor
```
idantene: Nitpicking, but why assign the class to a `const` instead of just declaring the class?
That is, if one still has to do `new NodeInterceptorConstructor(...)`, wouldn't it make more sense to have `new NodeInterceptor(...)`? I'd expect the former to be called once more to create a `NodeInterceptor`.

packages/unmock-core/src/backend/index.ts
ksaaskil: It's assigned below with `createInterceptor`. The idea is that each interceptor is only initialized once (when it's constructed). So if backend is initialized again, we just disable the old interceptor and remove reference to it.

packages/unmock-core/src/interceptor/index.ts
ksaaskil: Yes it's a future factory function, possibly dynamically creating a suitable interceptor (Node.js vs browser) based on the current environment.

packages/unmock-core/src/interceptor/node-interceptor.ts
ksaaskil: Yep the idea is simply that because creating the interceptor starts the interception (without a separate "initialize" call), I wanted to add typing to the constructor itself with `IInterceptorConstructor`. And AFAIK one cannot implement an interface with class constructor directly so the constructor is typed here by assigning the class. Maybe it's overengineering but I don't see harm in it.

ksaaskil: I mean this does not work:
```
interface IInterceptor {
  new (options: IInterceptorOptions): IInterceptor;
  disable(): void;
}
```
but one needs `IInterceptorConstructor`. And with the `const` assignment it can be made explicit that the `NodeInterceptor` static class implements `IInterceptorConstructor`.

packages/unmock-core/src/backend/index.ts
idantene: Ah, somehow missed the initializer there. My bad.

packages/unmock-core/src/interceptor/node-interceptor.ts
idantene: That's fine, I was mostly referring to the naming convention.
Calling `new NodeInterceptorConstructor(...)` IMO implies we're creating a new constructor, where as here it creates a new `NodeInterceptor`.

As for the interfaces - I don't mind much; I'm unaware of languages that allow interfaces for the constructor. I guess the more typical solution would be to use abstract classes though.

ksaaskil: Ah ok, I'll rename it in the next PR.

PR #309, opened 2019-10-09
Derive NodeBackend from abstract Backend class

ksaaskil:
- Create new abstract `Backend` class (it does not really have to be abstract as it only has one abstract method, `loadServices`)
- Bring back `IServiceDefLoader` and make `FsServiceDefLoader` to implement it to add decoupling
- After this, `UnmockPackage` should not directly depend on the Node.js environment

TODO:
- [ ]  The name `Backend` is a bit confusing, I think we could come up with a better name
- [ ]  There's still a lot to do if we want to isolate Node.js related stuff from non-Node.js


packages/unmock-core/src/backend/index.ts
```diff
@@ -124,17 +101,12 @@ export default class NodeBackend {
...
-        }),
-        FSSnapshotter.getOrUpdateSnapshotter({}),
-      ],
+      listeners: this.listeners,
       options,
       store: this.serviceStore,
     });
 
-    this.interceptor = createInterceptor({
+    this.interceptor = new this.InterceptorCls({
```
idantene: Not very critical, but it feels weird to call `createResponse` a listener. Is it the same kind of listener as `this.listeners`?

packages/unmock-core/src/service-loaders/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { IServiceDef } from "../interfaces";
+import { resolveUnmockDirectories } from "../utils";
+
+const debugLog = debug("unmock:node");
+
+export interface IServiceDefLoader {
+  loadSync(): IServiceDef[];
+}
+
+const createFsServiceDefLoader = (
```
idantene: Is this also intended as a future factory function?

packages/unmock-core/src/service-loaders/index.ts
```diff
@@ -0,0 +1,24 @@
+import debug from "debug";
+import * as _ from "lodash";
+import { FsServiceDefLoader } from "../fs-service-def-loader";
```
idantene: If we have `/service-loaders/`, maybe we can move `fs-service-def-loader` to this folder as well...

packages/unmock-core/src/backend/index.ts
ksaaskil: `listeners` was renamed to "onSerializedRequest" in #307 so I'll have to merge that here.

packages/unmock-core/src/service-loaders/index.ts
ksaaskil: Yep, and to hide some of the directory resolution.

packages/unmock-core/src/service-loaders/index.ts
ksaaskil: 👍

PR #311, opened 2019-10-11
Uses clearer language in the README

mikesol:
- Fixes passthrough, as it is a contresens as currently used
- Change the note on required query strings

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Mock API dependencies in JavaScript.
   + [Specifying replies](#specifying-replies)
   + [Specifying reply headers](#specifying-reply-headers)
   + [Chaining](#chaining)
-  + [Passthrough](#passthrough)
+  + [Trivial API calls](#trivial-api-calls)
 * [Expectations](#expectations)
 * [Initializing Mocks](#initializing-mocks)
 * [Runner](#runner)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,9 +250,9 @@ unmock. nock('http://myapp.iriscouch.com')
   })
 ```
 
-### Passthrough
+### Trivial API calls
 
-For "boring" API calls where you are just passing through information, you can instruct unmock to serve random `200` resposnes to those requests using `tldr`.
+For "trivial" API calls where you are just passing through information, you can instruct unmock to serve random `200` resposnes to those requests using `tldr`.
```
carolstran: A few notes about this section...

1) Can we remove `just`? I think the sentence works without it and this language is often considered condescending in documentation. We could also reword it to something like `For "trivial" API calls where the goal is to pass through information...` 🤔 

2) Typo --> `responses`

README.md
```diff
@@ -250,9 +250,9 @@ unmock. nock('http://myapp.iriscouch.com')
   })
 ```
 
-### Passthrough
+### Trivial API calls
```
carolstran: I also wonder if there's a better header for this... we can keep for the sake of this PR, but might revisit later.

mikesol: I just changed it to "Ignorable", which is perhaps not as easy for a non-native speaker as it is a neologism, but it may get the idea across better.

carolstran: Yeah maybe... let's sit on it for a while and if something else comes to mind we can always change it.

Merging now 🚀

PR #312, opened 2019-10-11
Add unmock-fetch.

ksaaskil:
- Adds `unmock-fetch` package for intercepting calls to `global.fetch` or `window.fetch`
- Also provides `buildFetch` that takes in a callback function of type `OnSerializedRequest`
- Used either via `const fetch = buildFetch(onSerializedRequest)` or with `UnmockFetch.on(onSerializedRequest); ... UnmockFetch.off();`. The latter patches `global.fetch` or `window.fetch`.
- Move `OnSerializedRequest` to `interfaces.ts` in `unmock-core` so it can imported

TODO:
- [ ]  Get rid of hackish imports from `unmock-core/dist/interfaces`. Importing from `unmock-core` is not possible at the moment as it imports a lot of Node.js related stuff and that would break in React Native


packages/unmock-fetch/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "type": "git",
+    "url": "git+https://github.com/unmock/unmock-js.git"
+  },
+  "homepage": "https://github.com/unmock/unmock-js",
+  "bugs": {
+    "url": "https://github.com/unmock/unmock-js/issues"
+  },
+  "dependencies": {
+    "@types/debug": "^4.1.5",
+    "@types/url-parse": "^1.4.3",
```
idantene: Should `debug` and its `@types` be part of the dependencies? (and not dev dependencies?)

packages/unmock-fetch/src/index.ts
```diff
@@ -0,0 +1,77 @@
...
+declare namespace global {
+  let fetch: Fetch;
+}
+
+// tslint:disable-next-line:no-namespace
+declare namespace window {
+  let fetch: Fetch;
+}
+
+let mitm: Mitm | undefined;
```
idantene: Is this the same `mitm` we have in the backend?

packages/unmock-fetch/package.json
ksaaskil: Yes, a library should include the types it uses.

ksaaskil:
Hmm, I don't see what you mean by "couldn't we have used `cross-fetch`"? Do you mean patching `crossFetch.fetch` instead of patching `global.fetch` or `window.fetch`? I tried something like that but could not figure out how to actually do it.

ksaaskil:
Or if you mean the `unmock.fetch` included in the package export, it would indeed probably be better if it was indeed set to `cross-fetch.fetch` instead of `node-fetch`.

packages/unmock-fetch/package.json
idantene: Of course, but `debug` is not a dependency. The code doesn't *need* it to function properly, so shouldn't it be part of dev dependencies?

packages/unmock-fetch/src/index.ts
idantene: I see that, but it seems you tried to maintain the same interface, hence the question. It's a bit misleading IMO to have two `Mitm` classes with different purposes.

idantene:
> Or if you mean the `unmock.fetch` included in the package export, it would indeed probably be better if it was indeed set to `cross-fetch` instead of `node-fetch`.

Yup, meant this :)

packages/unmock-fetch/src/index.ts
ksaaskil: Yeah I can rename it if it makes development simpler for us. But in general users of this would not know anything about `node-mitm` and its API so I just picked a short name that makes it clear what it does. I can call it `FetchInterceptor` or something.

packages/unmock-fetch/package.json
ksaaskil: But `debug` is a dependency of `unmock-fetch` :) It's true that user's type-checking probably would pass fine even if `@types/debug` wasn't part of the dependencies, but it's generally just bad practice to leave out the types. We've had two breakages in releases already because of leaving out types for `sinon` and `node-fetch`, requiring users to manually add the types.

PR #313, opened 2019-10-11
Extract unmock-node from unmock-core.

ksaaskil:
Here comes a hugey!
- Create new package `unmock-node` in preparation for `unmock-browser`, `unmock-react-native`, etc.
- Move all NodeBackend-related stuff to `unmock-node`, including lots of end-to-end tests
- Remove `CustomConsole` as it's not even used at the moment
- Move `src/index.ts` to `src/node/index.ts` in `unmock` to prepare for multiple builds for different environments
- Exports are not handled very prettily at the moment, `unmock-core` exports a **lot** of stuff that's needed by other packages, and the top-level `unmock` package exports everything that `unmock-core` exports. I think it's fine for `unmock-core` to export all kinds of crap, but `unmock` should only export selectively.

TODO in another PR:
- [ ]  Clean the top-level exports from `unmock` package by removing `export * from "unmock-core"` 
- [ ]  Try to get `unmock-core` and `openapi-refinements` to actually work in the browser and React Native, currently crashes (at least) because `jsonschema` package depends on Node stuff


packages/unmock-core/src/index.ts
```diff
@@ -1,24 +1,25 @@
...
-import FsSnapshotter, { ISnapshot } from "./loggers/snapshotter";
-import WinstonLogger from "./loggers/winston-logger";
+import { ISnapshot } from "./interfaces";
 import { ExtendedJSONSchema, nockify, vanillaJSONSchemify } from "./nock";
 import internalRunner, { IRunnerOptions } from "./runner";
+import { IService } from "./service/interfaces";
 import { AllowedHosts, BooleanSetting } from "./settings";
 
-export { fetch };
+export * from "./interfaces";
```
ksaaskil: This is bad, maybe we like to create `unmock-types` just containing the core types so that libraries like `unmock-fetch` can depend only on that.

ksaaskil:
Wow code coverage takes a serious hit even though this PR does not add or remove any tests, just re-organizes them. The reason is probably that a lot of the core functionalities are now only tested `unmock-node` so they aren't counted. One needs to add more tests to unmock-core but I'd rather do that later on.

packages/unmock-core/src/console/index.ts
```diff
@@ -1,7 +1,2 @@
-import { CustomConsole } from "./console";
-
-const customConsole = new CustomConsole(process.stdout, process.stderr);
-
-export { CustomConsole, formatMsg } from "./console";
+export { formatMsg } from "./console";
 export { Chalks } from "./chalks";
```
idantene: Is `Chalks` still used anywhere?

packages/unmock-core/src/index.ts
```diff
@@ -1,24 +1,21 @@
...
 export * from "./types";
+export { ISnapshot } from "./interfaces";
 export { sinon };
 export { u } from "./nock";
 export { transform, Addl, Arr } from "./generator-utils";
-
-export { ISnapshot };
-const utils = { snapshotter: FsSnapshotter };
-export { utils };
+export { isRESTMethod } from "./interfaces";
```
idantene: This and `ISnapshot` are exported with the `export * from ./interfaces`

packages/unmock-node/src/__tests__/backend/index.test.ts
```diff
@@ -1,8 +1,8 @@
 import axios from "axios";
 import { removeCodes } from "openapi-refinements";
 import * as path from "path";
-import { Service, sinon, transform, UnmockPackage } from "../../";
-import NodeBackend from "../../backend/node";
+import { Service, sinon, transform, UnmockPackage } from "unmock-core";
```
idantene: Obviously not for this PR, but maybe we can retire `UnmockPackage`? :thinking:

packages/unmock-node/src/fs-service-def-loader.ts
```diff
@@ -1,9 +1,10 @@
 import debug from "debug";
 import * as fs from "fs";
+import * as _ from "lodash";
```
idantene: where is `lodash` used here?

packages/unmock/src/node/index.ts
```diff
@@ -1,4 +1,4 @@
-import unmock from "unmock-core";
-
 export * from "unmock-core";
+import unmock from "unmock-node";
+export * from "unmock-node";
 export default unmock;
```
idantene: Do we want `unmock-node` to be the default?

packages/unmock-node/src/__tests__/backend/index.test.ts
idantene: Or well, more like rename it? Or... yeah not sure, actually maybe since we're planning for a browser version and such, it might finally earn its place..?

packages/unmock-core/src/index.ts
idantene: Agreed, even exporting it as `unmock-core/interfaces` as a first step would be cleaner.
Definitely for a different PR though...

packages/unmock-core/src/console/index.ts
ksaaskil: Not sure but I think it's logical for the `console` module to export it.

packages/unmock-core/src/index.ts
ksaaskil: 👍

packages/unmock-node/src/__tests__/backend/index.test.ts
ksaaskil: Yep I think it's fine for it to exist as the top-level API, ensuring all packages export an instance of the same class.

packages/unmock/src/node/index.ts
ksaaskil: Hmm I don't quite understand the question, the default import of `unmock-node` is here set as the default import of `unmock`.

packages/unmock-node/src/fs-service-def-loader.ts
ksaaskil: Removed, linter was confused as comments included underscore.

packages/unmock/src/node/index.ts
idantene: My bad, didn't notice the path (thought it's `packages/unmock/index.ts`)

PR #315, opened 2019-10-14
Add more tests to core.

ksaaskil:
Copy core tests from `dynamicService.test.ts` to `nock.test.ts` in core, skipping a few that were more end-to-end and need an interceptor.

- [x]  Should the duplicate tests from `unmock-node` be removed?


packages/unmock-node/src/__tests__/dynamicService.test.ts
```diff
@@ -314,129 +35,6 @@ describe("Tests dynamic path tests", () => {
...
-        .reply(200, "", { hello: "world" });
-      expectNServices(1);
-      expect(
-        getPrivateSchema("foo").paths["/"].get.responses["200"].headers.hello
-          .schema,
-      ).toEqual({ type: "string", enum: ["world"] });
-    });
-  });
-
   describe("schema generates valid stuff", () => {
```
carolstran: I know this wasn't part of your changes for this PR, but can we use a more descriptive word than "stuff" here? 😂

packages/unmock-core/src/__tests__/corePackage.test.ts
```diff
@@ -1,21 +1,5 @@
...
-  }
-  public reset() {
-    return;
-  }
-  public get services() {
-    return {};
-  }
-  public loadServices() {}
-}
+import { TestBackend } from "./utils";
```
carolstran: Smart move 👍

packages/unmock-core/src/__tests__/corePackage.test.ts
```diff
@@ -1,21 +1,5 @@
 import { UnmockPackage } from "..";
```
carolstran: Is there a reason you all decided to use relative paths rather than absolute?

packages/unmock-core/src/__tests__/nock.test.ts
```diff
@@ -0,0 +1,435 @@
...
+
+    it("A reply header is correctly propagated", () => {
+      expectNServices(0);
+      unmock
+        .nock("https://www.foo.com", "foo")
+        .get("/")
+        .reply(200, "", { hello: "world" });
+      expectNServices(1);
+      expect(
+        getPrivateSchema("foo").paths["/"].get.responses["200"].headers.hello
```
carolstran: I'm wondering if we can use more meaningful placeholders than "foo", "bar", etc. 🤔 That way, the tests better explain the functionality and maybe even enable onboarding into the project.

This doesn't need to be part of this PR as you were moving these tests, not writing them from scratch - but something to consider.

packages/unmock-node/src/__tests__/dynamicService.test.ts
ksaaskil: We should! I fully agree 🙂

packages/unmock-core/src/__tests__/corePackage.test.ts
ksaaskil: You mean absolute like `import { UnmockPackage } from "unmock-core"`? That would work as well, but wouldn't that depend on the `moduleNameMapper` in `jest.config.js` to be configured to point to the source directory? The relative path makes it imho easier to see that one's importing from the same package.

packages/unmock-core/src/__tests__/nock.test.ts
ksaaskil: Definitely!

PR #316, opened 2019-10-14
Replace IInterceptorConstructor with IInterceptorFactory.

ksaaskil:
It's easier to mock a factory function creating interceptors with `jest.fn` than to mock a constructor type. Also storing a class like `InterceptorCls` as an instance property just feels icky  🤒


carolstran:
Out of curiosity, why does it feel icky @ksaaskil? Not disagreeing, but trying to gain more context before reviewing 😇

ksaaskil:
> Out of curiosity, why does it feel icky @ksaaskil? Not disagreeing, but trying to gain more context before reviewing 😇

It just feels really object-oriented, as the interceptor basically needs to be a concrete "class" that can be instantiated with `new`. And it's simpler imho to mock the dependency if it's just a function that returns an object and not a constructor: if I wanted to create an interceptor that wraps two interceptors, I'd have to basically create a `TwoInterceptorWrapper` class.

PR #318, opened 2019-10-16
Unmock server

ksaaskil:
## Changes

- Adds `unmock-server` package
- Contains a proxy (`proxy.ts`) and unmock server (`server.ts`)
- Proxy receives requests and forwards them to the unmock server. **Note that the proxy cannot look at the traffic in the case of HTTPS request, at least not without decoding with the private key. Proxy should maybe rather send back its own certificate and communicate with the target host directly.**
- Server receives proxied requests and looks at `X-Forwarded-Host` (or `Host`) header to determine where the request was going and serializes the request to a `ISerializedRequest`
- `ISerializedRequest` is handled with `unmock`

## Instructions

1. Create self-signed certificate (`key.pem` and `cert.pem`) for a given domain:
    > `bash packages/unmock-server/scripts/prepare-cert.sh petstore.swagger.io` 
    > ~`openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes -subj '/CN=*.swagger.io'`~
1. Create `__unmock__` folder with service for `petstore.swagger.io/v2`
1. Start server: `npm run compile && node packages/unmock-server/index.js`. This starts the proxy at `localhost:8008` and Unmock server at `localhost:8000` (HTTP) and `localhost:8443` (HTTPS)
1. Make request using the proxy, trusting the certs: `http_proxy=http://localhost:8008 HTTPS_PROXY=http://localhost:8008 SSL_CERT_FILE=cert.pem curl -i http://petstore.swagger.io/v2/pet/23`

### Certificates in case of HTTPS

1. The certificate used by the Unmock server must be explicitly trusted by the client.
1. Alternatively, the client must accept invalid certificates (á la `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'`)
1. How to create a certificate valid for all the domains mocked by the server? Should they be created on-fly when creating the service and bundled together? How does [Hoverfly use their certificate](https://docs.hoverfly.io/en/latest/pages/tutorials/basic/https/https.html)? 

### Setting the proxy

- Some applications and libraries respect `http_proxy` and `https_proxy` (or uppercased) environment variables, so if application is using those, requests are easy to proxy to the running Unmock proxy
- Node.js does not respect the environment variable (https://github.com/nodejs/node/issues/8381). (But in Node.js, one can just use `unmock` library)
- Some tools like [proxychains](https://github.com/haad/proxychains) might be useful to look at
- If we build our own Docker images, [they can be set to use a proxy](https://docs.docker.com/network/proxy/)

### Managing service state

- PR contains a primitive API at `/api` for setting state
- How should this work in general?


package.json
```diff
@@ -94,6 +95,7 @@
     "unmock-fetch": "file:packages/unmock-fetch",
     "unmock-jest": "file:packages/unmock-jest",
     "unmock-node": "file:packages/unmock-node",
+    "unmock-server": "file:packages/unmock-server",
```
MeeshkanCain: Should this also be included in the `lint-ts` and `lint-ts-fix` scripts above?

packages/unmock-server/src/proxy.ts
```diff
@@ -0,0 +1,95 @@
...
+  proxySocket.on("data", chunk => {
+    socket.write(chunk);
+  });
+
+  proxySocket.on("end", () => {
+    socket.end();
+  });
+
+  proxySocket.on("error", () => {
+    socket.write("HTTP/" + req.httpVersion + " 500 Connection error\r\n\r\n");
```
MeeshkanCain: Could create a function to maintain the `HTTP/${req.httpVersion} ${msg}\r\n\r\n` pattern.

packages/unmock-server/src/proxy.ts
```diff
@@ -0,0 +1,95 @@
...
+    socket.end();
+  });
+
+  socket.on("data", chunk => {
+    proxySocket.write(chunk);
```
MeeshkanCain: Do we need any checks on the chunk being written (e.g. that it ends with `\r\n\r\n`)?

packages/unmock-server/src/server.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import {
+  ISerializedRequest,
+  ISerializedResponse,
+  transform,
+} from "unmock-core";
+import { createUnmockAlgo } from "./unmock";
+
+const httpPort = 8000;
+const httpsPort = 8443;
```
MeeshkanCain: This is linked to the port defined in `proxy.ts`; could benefit from a shared definition to avoid potential breakage.

packages/unmock-server/src/server.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  app.all("*", handler);
+
+  return { unmock, app };
+};
+
+export const startServer = (app: express.Express) => {
+  const options = {
+    key: fs.readFileSync(`${process.cwd()}/key.pem`),
+    cert: fs.readFileSync(`${process.cwd()}/cert.pem`),
```
MeeshkanCain: I'd say a `.env` file could be more useful here than `cwd`

package.json
ksaaskil: 👍

packages/unmock-server/src/proxy.ts
ksaaskil: I'm not sure we need that, it's up to the target server to return proper data. Also not sure how that would work with encryption.

packages/unmock-server/src/server.ts
ksaaskil: Probably, will do later on if we work on this more.

packages/unmock-server/src/server.ts
ksaaskil: Added `constants.ts`.

packages/unmock-server/src/proxy.ts
ksaaskil: 👍

PR #319, opened 2019-10-16
Refactor: Make Backend concrete, expose algorithm.

ksaaskil:
- Make `Backend` a concrete class by adding `serviceDefLoader` as dependency
  - Gets rid of inheritance 🎉 
  - Puts all "backend" logic into one place
  - Classes like `NodeBackend` are then just wrappers calling `super`
- Expose `backend` and `backend.handleRequest` from `unmock` object. 
  - Users of `UnmockPackage` can then access the algorithm without hacking the interceptor, like in #318

### DIFF
--- a/packages/unmock-node/src/__tests__/dynamicService.test.ts
+++ b/packages/unmock-node/src/__tests__/dynamicService.test.ts
@@ -2,8 +2,7 @@ import axios from "axios";
 import { u } from "unmock-core";
 import unmock from "..";
 
-const expectNServices = (expectedLength: number) =>
-  expect(Object.keys(unmock.services).length).toEqual(expectedLength);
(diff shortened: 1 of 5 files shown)


packages/unmock-node/src/__tests__/dynamicService.test.ts
```diff
@@ -12,12 +11,12 @@ describe("Tests dynamic path tests", () => {
   beforeEach(() => unmock.reloadServices());
   describe("Paths can be defined using arrays and regexs", () => {
     it("Also handles multiple parameters", async () => {
-      expectNServices(0);
+      expect(nServices()).toEqual(0);
```
ksaaskil: Changed this so that when the test fails, correct line is shown.

PR #320, opened 2019-10-16
Add runner tests to core.

ksaaskil:
- Depend on `unmock-fetch` to do simple "end-to-end" testing with HTTP requests without depending on `node-mitm`, increases coverage in the core
- Fix a bug in `unmock-fetch` where the query string was not parsed
- Keeps duplicate `runner.test.ts` in `unmock-node` as they depend on `node-mitm` so they don't test the same thing, kinda
- [x]  Includes #319, rebase once merged
- [ ]  A less hacky and more portable alternative might be to add `fetch` in `UnmockPackage`, so that all requests made via `unmock.fetch` in tests would be handled by unmock (after `unmock.on()` without using an interceptor


packages/unmock-core/src/__tests__/runner.test.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const backend = new Backend({
+  interceptorFactory,
+  serviceDefLoader: testServiceDefLoader,
+});
+
+const unmock = new UnmockPackage(backend);
+
+const runner = unmock.runner.bind(unmock);
```
idantene: Does one have to use `unmock.runner.bind(unmock)` now? Or is it simply because we're testing this within the `core` package?

ksaaskil: Used simply `unmock.runner` instead.

PR #328, opened 2019-10-18
Simplify fetch interface.

ksaaskil:
- Simplify the fetch interface: allow passing in a function of type `CreateResponse: (req: ISerializedRequest) => ISerializedResponse | undefined`, not unlike in the Python version
- It would be nicer if `CreateResponse` was typed so that the response is `ISerializedResponse` and never `undefined`. This goes back to the old discussion of whether the algorithm should return `undefined` or just throw an exception when no match is found. I'm maybe starting to change my mind, not finding a matcher should just be an exception. Maybe an exception of specific instance still so that the core knows to handle it differently than unexpected errors, if needed.

### DIFF
--- a/packages/unmock-fetch/README.md
+++ b/packages/unmock-fetch/README.md
@@ -20,34 +20,23 @@ yarn add unmock-fetch -D
 
 ```ts
 import { buildFetch } from "unmock-fetch";
-import { ISerializedRequest, ISerializedResponse, OnSerializedRequest } from "unmock-core";
+import { CreateResponse, ISerializedRequest, ISerializedResponse, OnSerializedRequest } from "unmock-core";
 
 // Define what to do with the intercepted request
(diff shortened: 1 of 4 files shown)


ksaaskil:
Any comments to this? @idantene

idantene:
Woops! Sorry, forgot about that for some reason. LGTM; personally
`cb.length` is a bit weird - I guess it refers to number of arguments?
Maybe a comment will be useful (i.e. number of arguments for CreateResponse
and OnSerializedRequest must not be equal?)

ksaaskil:
Yep I agree the check is not very clean, added a comment. I'd maybe in future remove the `OnSerializedRequest` interface completely, that's just how the interceptors work at the moment (to avoid replicating too much of "algorithmic" logic) so it's good to keep for now.

PR #336, opened 2019-10-28
Add setting for randomization.

ksaaskil:
- Currently Unmock always sets the seed to a fixed value when generating a mock. For proper pseudo-random behaviour, add a `randomize` setting to not set the seed everytime.
- Switching randomization on: `unmock.randomize.on()`
- Disable: `unmock.randomize.off()`

TODO:
- [x]  Tests
- [ ]  Completely based on mutable state: having RNG based on functional immutable state (á la [fp-ts/lib/state.ts](https://github.com/gcanti/fp-ts/blob/master/src/State.ts)) would be awesome but I don't really see that fitting to the current architecture at the moment: We would need to change `CreateResponse` to return the modified state.

### DIFF
--- a/packages/unmock-core/src/generator.ts
+++ b/packages/unmock-core/src/generator.ts
@@ -61,9 +61,11 @@ import {
 import { ServiceStore } from "./service/serviceStore";
 
 export const runnerConfiguration = {
+  defaultSeed: 0,
   optionalsProbability: 1.0,
   minItems: 0,
   reset() {
+    this.defaultSeed = 0;
     this.minItems = 0;
     this.optionalsProbability = 1.0;
   },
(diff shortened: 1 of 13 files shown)


ksaaskil:
Let me actually move `rng` to `Backend`, I think it fits much better there...

packages/unmock-core/src/generator.ts
```diff
@@ -59,14 +59,15 @@ import {
...
 } from "./service/interfaces";
 import { ServiceStore } from "./service/serviceStore";
+
 export const runnerConfiguration = {
-  seed: 0,
+  defaultSeed: 0,
   optionalsProbability: 1.0,
   minItems: 0,
   reset() {
+    this.defaultSeed = 0;
```
idantene: If we now use `defaultSeed` as an immutable instance, why do we need to reset it here?

ksaaskil:
Good points, I added `rng.restore()` to abstract the default seed away.

PR #340, opened 2019-10-31
Add code for self-signing certificates on the fly.

ksaaskil:
- Circumvents the requirement to generate certificates matching mocked domains before-hand
- Generate a certificate on the fly for given request (should probably be memoized)
- Working with HTTPS then requires that clients trust `ca.pem` included in the repository (similarly as for [Hoverfly](https://hoverfly.readthedocs.io/en/v0.10.0/pages/tutorials/https/https.html)).
- Adds script `openssl-gen.sh` for generating CA certificate and key ([config](https://github.com/unmock/unmock-js/blob/snicallback/packages/unmock-server/certs/openssl.cnf) copied from [goproxy](https://github.com/elazarl/goproxy/blob/master/certs/openssl.cnf), changed to refer to unmock).
- Includes the key and certificate in the repository like [Hoverfly does](https://github.com/SpectoLabs/hoverfly/blob/master/core/cmd/hoverfly/main.go)

TODO:
- [ ]  Depends on client supporting SNI (TLS extension). Is this the case practically always?
- [x]  Add public and private keys for the CA to repository
- [ ]  Update instructions
- [x]  Tests
- [ ]  It would be more secure to generate a new CA for every client (like [mitmproxy](https://docs.mitmproxy.org/stable/concepts-certificates/) does). This would require the user to first generate the certificate, trust it, and then run the tests.

Affected repositories:
- [ ] [unmock-server-docker](https://github.com/unmock/unmock-server-docker/pull/1)


ksaaskil:
That's right, clients need to trust `ca.pem` included in this repository. I'm not sure I would call that a "problem", it's an inconvenience that's a built-in security feature in HTTPS.

mikesol:
Great, got it. Thanks for the clarification – I know very little about this, so I probably have a lot of assumptions baked into my thinking that I need to unwind/undo.
In any case LGTM!

PR #355, opened 2019-12-12
Extract UnmockFaker from Backend.

ksaaskil:
- Extract `UnmockFaker` class responsible for faking responses to serialized requests, lifting that responsibility from `Backend` class.
- Refactor `unmock.nock` so that the same API can be used also from `UnmockFaker` to add services
- Add helper methods to `ServiceStore`: `update`, `removeAll`, `resetServices`

The purpose of all this is that one can separate the mock generation logic from the interceptor logic: one can create mock responses via `UnmockFaker` without starting the interceptor and add new services and modify their states. And it's nice that one can create multiple independent fakers, as the big bad global interceptor is separate from the faking logic.

### DIFF
--- a/packages/unmock-core/src/__tests__/faker.test.ts
+++ b/packages/unmock-core/src/__tests__/faker.test.ts
@@ -1,0 +1,40 @@
+import { u } from "..";
+import UnmockFaker from "../faker";
+import { ServiceStore } from "../service/serviceStore";
+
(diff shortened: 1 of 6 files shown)


ksaaskil:
Yeah we should add something about it in the documentation! It's currently however more for internal development and it'll likely change quite a bit as new needs emerge, so I wouldn't maybe do that yet.

PR #358, opened 2019-12-14
Create unmock-types package

### LINKED ISSUE #357: Extract types from unmock-core to a new package
It would be very useful to extract common types from [interfaces.ts](https://github.com/unmock/unmock-js/blob/dev/packages/unmock-core/src/interfaces.ts) to a separate `lerna` package called, for example, `unmock-types`. Once this is done, one could remove for example `unmock-core` dependency from `unmock-fetch` and only depend on `unmock-types`. Types could then also be used in other projects.

What should be moved:
- `ISerializedRequest` and `ISerialzedResponse` and everything they depend on
- `OnSerializedRequest` and `CreateResponse`

How to create a new package:

1. Copy an existing (light-weight) package such as `unmock-jest` to a new package called `unmock-types`
1. Update `package.json`
1. Update `tsconfig.json` and its `references` in the new package
1. Remove everything from `src/` except `index.ts` and `src/__tests__/tsconfig.json`
1. Update [tsconfig.build.json](https://github.com/unmock/unmock-js/blob/dev/tsconfig.build.json) in the project root to refer to the package
1. Update [tsconfig.test.json](https://github.com/unmock/unmock-js/blob/dev/tsconfig.test.json)
1. Add the package as a dependency in the root [package.json](https://github.com/unmock/unmock-js/blob/dev/package.json#L91)
1. Add the package to `lint-ts` and `lint-ts-fix` commands in `package.json`

Instead of copying, you can also use the `lerna` CLI's [create](https://github.com/lerna/lerna/tree/master/commands/create#readme) command to create the package.

For this issue, it's not required to update other packages to use the new package. That can (and should) be done in later PRs.

k4m4:
Fixes #357

## Description

Creates a separate `lerna` package called `unmock-types` that contains commonly used types (from [interfaces.ts](https://github.com/unmock/unmock-js/blob/dev/packages/unmock-core/src/interfaces.ts)).


k4m4:
@carolstran I totally agree. Although, in this particular example, I am not entirely sure what else we could include in the docs. This package just contains commonly used types, which is what is currently denoted in the readme. Any particular ideas, before we merge this?

carolstran:
@k4m4 Oh sorry, I should've been more clear - my comment shouldn't block merging! It was more of a general note, but I'm planning on opening some documentation-related issues later so we don't need to address it in this PR.

Merge away 🚀

k4m4:
@carolstran Got it, thanks!

ksaaskil:
The package is now available in [npm](https://www.npmjs.com/package/unmock-types)! I'll probably use those only in [diglett](https://github.com/Meeshkan/diglett) in the short run but we can slowly work on migrating also `unmock-js` to use the package. I'm planning to write open up discussion how the whole `unmock-js` could be refactored to be more modular, so that'll probably involve some planning what should be included in `unmock-types` (like `IInterceptor`, `IFaker` etc.)

PR #359, opened 2019-12-16
Polyfill Buffer when missing from global.

ksaaskil:
Getting `unmock` run in React Native (https://github.com/unmock/unmock-react-native-example/pull/3) requires polyfilling `Buffer` as we depend on that somewhere via `whatwg-url`. Messing with `global` is not nice at all but I think this is the easiest solution for now to get browser support move forward.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3591,6 +3591,11 @@
         }
       }
     },
+    "base64-js": {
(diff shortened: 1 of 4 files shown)


ksaaskil:
OK to merge @carolstran ?

PR #364, opened 2019-12-19
Release v0.3.11

carolstran:
Features:

  - Add CLI to `unmock-server`
  - Add `unmock-browser` package, which runs `unmock-js` in browser and React Native
  - Implement parsing of request headers in `unmock-fetch`
  - Add `unmock-types` package
  - Extract `UnmockFaker` class from `Backend` class
  - Rename `unmock.newFaker()` to `unmock.faker()`

Fixes:

  - Remove `concurrently` dependency


changelog.md
```diff
@@ -1,5 +1,24 @@
 # CHANGELOG.md
 
+## 0.3.11 (2019-12-19)
+
+Breaking changes:
+
+  - Rename `fake()` to `generate()`
+  - Rename `unmock.newFaker()` to `unmock.faker()`
```
ksaaskil: These methods were added after the previous release so `fake` and `unmock.newFaker()` have not been part of any release, so nothing breaking 😇

carolstran: Should I move them to the Features section then or rather remove them entirely?

ksaaskil: I think the one about `unmock.faker()` could be added to features after `UnmockFaker`, the other one can be removed. But it's not that important, not sure if anyone reads change logs anyway 🙂

Issue #299, opened 2019-10-07
Mocha support


mikesol:
Currently, unmock hardcodes a lot of conventions for jest and will break on mocha, especially the runner. Also, the `jest-reporter` will not work in Mocha. Everything should work in mocha as it does in Jest, with additional tests in mocha (probably in end-to-end) to make sure this is the case.

carolstran:
Dropping this here to track progress!

TODO:
- ✅ Extract runner from unmock-core (#380)
- ✅ Release version 0.3.13 of unmock-js (#381)
- ✅ Build an `unmock-jest-runner` package that will replace the `jestRunner` ([#1](https://github.com/Meeshkan/unmock-jest-runner/pull/1) on [unmock-jest-runner](https://github.com/Meeshkan/unmock-jest-runner))
- Release a `0.0.0` version on npm
- Replace instances of `jestRunner` in `unmock-js` with new `unmock-jest-runner` package
- Release a version 0.3.14 of `unmock-js`
- Update the documentation, README.md, koans, unmock-examples and unmock-katacoda to use `unmock-jest-runner`
- Build an `unmock-mocha-runner` package. "This is the one that IMO is the most perilous, as we have not yet tested if the abstraction in `unmock-runner` is powerful enough to handle what mocha needs. For example, mocha may have some error validation scheme or return values from tests that require the abstraction in `unmock-runner` (ie the types) to change. In the ideal case, though, it'll be as short as the `jestRunner` and everything will work." - @mikesol

PR #380, opened 2020-02-10
Extract runner from unmock-core

carolstran:
Part of #299 

- Created separate `unmock-runner` package
- Made exports more generic to be able to run with Jest, Mocha or any other future imported runner function
- Added a temporary `jestRunner` function under `unmock-runner` to use until we have a proper `unmock-jest-runner` package

TODO in future PRs: 
- Publish this as a package on NPM 
- Create/publish `unmock-jest-runner` (to replace `packages/unmock-runner/src/jestRunner`) 
- Create/publish `unmock-mocha-runner`
- Add appropriate documentation and configuration (ie add unmock-runner to `lint-ts` and `lint-ts-fix` commands)

@mikesol My build is still failing because it cannot find the `unmock-runner` module. Assuming this is a lerna thing? Any tips for how to resolve would be appreciated.


carolstran:
Makes sense @mikesol, although I also think it's something that will soon be replaced (as in the next day or two) so maybe not urgent for the time being.

Any thoughts on the build error though? Because that is blocking for merge.

mikesol:
Agreed that it is bikeshedding.
The most recent build I can see is 3260, and in there, the issue is this line:

```typescript
import runner, { IMeeshkanDoneCallback, IRunnerOptions } from "unmock-runner";
```

I think it should be this instead (haven't tested...):

```typescript
import runner, { IMeeshkanDoneCallback, IRunnerOptions } from "./";
```

carolstran:
Everything's passing now! @mikesol, let me know if I should merge.

mikesol:
Cool! @ksaaskil can also add his review, after which my recommendation would be the following:

- merge
- release
- build an `unmock-jest-runner` package that is the `jestRunner`, test and release a `0.0.0` version on npm
- delete the `jestRunner` and release a new `unmock-js`
- update the documentation, README.md, koans, unmock-examples and unmock-katacoda to use `unmock-jest-runner`
- build an `unmock-mocha-runner` package. this is the one that IMO is the most perilous, as we have not yet tested if the abstraction in `unmock-runner` is powerful enough to handle what mocha needs. For example, mocha may have some error validation scheme or return values from tests that require the abstraction in `unmock-runner` (ie the types) to change. In the ideal case, though, it'll be as short as the `jestRunner` and everything will work.

packages/unmock-node/src/__tests__/runner.test.ts
```diff
@@ -1,16 +1,17 @@
 import axios from "axios";
 import * as path from "path";
 import { Service, transform, UnmockPackage } from "unmock-core";
-import { runner } from "..";
 import NodeBackend from "../backend";
 const { withCodes } = transform;
+import jestRunner from "../../../unmock-runner/src/jestRunner";
```
ksaaskil: Hmm, I don't think one should refer to other Lerna packages via "../../../" but with something like `import { runner} from "unmock-runner"`. Jest will figure out that `unmock-runner` should point to the local package.

carolstran:
@mikesol Will the `unmock-runner` be accessible via the `unmock-js` package or do we need to create its own npm package? Asking because some of the types/initial setup from the runner is necessary for `unmock-jest-runner`

@ksaaskil That's a good point and I do agree, but it's going to be replaced as soon as we implement the `unmock-jest-runner` (hopefully today or tomorrow). However, if it really bothers you - I can fix it now!

ksaaskil:
> @mikesol Will the `unmock-runner` be accessible via the `unmock-js` package or do we need to create its own npm package? Asking because some of the types/initial setup from the runner is necessary for `unmock-jest-runner`
> 
> @ksaaskil That's a good point and I do agree, but it's going to be replaced as soon as we implement the `unmock-jest-runner` (hopefully today or tomorrow). However, if it really bothers you - I can fix it now!

No reason to fix it now, it was just a comment :)

mikesol:
> @mikesol Will the unmock-runner be accessible via the unmock-js package or do we need to create its own npm package? Asking because some of the types/initial setup from the runner is necessary for unmock-jest-runner
> @ksaaskil That's a good point and I do agree, but it's going to be replaced as soon as we implement the unmock-jest-runner (hopefully today or tomorrow). However, if it really bothers you - I can fix it now!

`unmock-runner` will be its own `npm` package that we'll import from, and it will be published automatically by `lerna`. The types will be accessible from that package. `unmock-js` is not a package AFAIK. Does that answer your question?

carolstran:
@mikesol Yes it does, thank you! Didn't realize that lerna publishes packages automatically.

PR #385, opened 2020-02-12
Release/v0.3.14

carolstran:
Fixes:

  - Fix import type error that made Unmock fail to compile


carolstran:
Oh that's smart, I hadn't considered that! I'll try that with this release, thanks @ksaaskil 🙏

PR #388, opened 2020-02-12
Replace jestRunner with unmock-jest-runner

carolstran:
Removed `jestRunner.ts` from `unmock-runner` package in exchange for the new [`unmock-jest-runner`](https://github.com/Meeshkan/unmock-jest-runner) package.

Part of #299.


packages/unmock-runner/package.json
```diff
@@ -14,5 +14,8 @@
   },
   "dependencies": {
     "unmock-core": "file:../unmock-core"
+  },
+  "devDependencies": {
+    "unmock-jest-runner": "0.0.0"
```
ksaaskil: Hmm I'm not sure if packages should add their own devdependencies like this, I think we've usually kept them at the root level `package.json`. One never runs tests for a single package but for the whole Lerna project. Could this maybe be removed and added to root-level dev dependencies?

packages/unmock-node/package.json
```diff
@@ -25,6 +25,7 @@
     "readable-stream": "^3.4.0",
     "shimmer": "^1.2.1",
     "unmock-core": "file:../unmock-core",
+    "unmock-jest-runner": "0.0.0",
```
ksaaskil: Should the node package ship with `unmock-jest-runner`?

packages/unmock-runner/package.json
carolstran: Ah that's a good point about lerna that I hadn't considered. Thank you! I can update 😊

packages/unmock-node/package.json
carolstran: Based on your previous comment... probably not 😅 I'll adjust!

carolstran:
So I added `unmock-jest-runner` to the root level but when I try to uninstall it from `unmock-node` and `unmock-runner`, I get a variation of this error and I'm not sure how to resolve it... 🤔 Have either of you seen something like this before? @ksaaskil @mikesol 

Screen Shot 2020-02-14 at 11 41 23 PM

mikesol:
It looks like something in the background has done `chmod` on the `openapi-refinements` directory. I find that the best way to resolve this is to nuke the repo and clone anew, as otherwise you'll need to do `chmod` calls by hand on directories, which can get dicey.

ksaaskil:
> So I added `unmock-jest-runner` to the root level but when I try to uninstall it from `unmock-node` and `unmock-runner`, I get a variation of this error and I'm not sure how to resolve it... 🤔 Have either of you seen something like this before? @ksaaskil @mikesol
> 
> Screen Shot 2020-02-14 at 11 41 23 PM

Lerna is tricky: you're not supposed to do `npm install` or `npm uninstall` inside packages as it will mess up how Lerna works. The way to add a dependency to a single package is either `npx lerna add my-package --scope=unmock-node` or adding the dependency by hand in the package's `package.json` and doing fresh install. If you want to remove a dependency, you need to delete it by hand from `package.json` and do a fresh install so that `package-lock.json` updates. I usually do `rm -rf node_modules && npm run clean && npm install` for a full fresh install.

packages/unmock/package.json
```diff
@@ -20,6 +20,7 @@
     "unmock-browser": "file:../unmock-browser",
     "unmock-cli": "file:../unmock-cli",
     "unmock-core": "file:../unmock-core",
+    "unmock-jest-runner": "0.0.0",
```
ksaaskil: I think the `unmock` package should also not include the jest runner if it's supposed to something people can install separately?

ksaaskil:
Hmm, I see there are now multiple `package-lock.json` files inside packages. We haven't added those to individual packages and I'm not sure if packages should even have that in Lerna... Maybe they should, but for the safe side, I'd consider deleting those for now.

carolstran:
What a mess ahhhhhh 🙈  But thank you both for your comments! Really appreciate it as this is my first time working with a Lerna project. I'll tackle this tomorrow 😊

PR #389, opened 2020-02-14
Does not throw on broken format

mikesol:
Fixes the fact that unmock cannot load a spec with a broken format.

## Description

Tells json schema faker not to throw if there is a broken format.

## How to test 

One of the `petstore` yml files has been modified to include a broken format.


mikesol:
Apologies, I should have been more clear. It's because in OpenAPI format can be anything, whereas in `json-schema` it is an enum of fixed values (`int32`, `int64` for `{"type":"integer"}`, for example).  As a result, when we pass an OpenAPI schema to `json-schema-faker`, there is a danger that a custom format will throw. This happened to me, for example, with the Stripe API, which has a custom format of "decimal" that caused unmock to crash.

PR #390, opened 2020-02-14
Docs: Update runner guide to include unmock-jest-runner

carolstran:
Blocked by #391 

Added a description and some usage details to the root README and the README for the `unmock-runner` package.

Some questions still lingering:
- Is it possible to change the default?
- Do we need the same documentation on both READMEs or would it be better to have the full documentation within the `unmock-runner` and then give a shorter description with a link on the root README?

Related to https://github.com/Meeshkan/unmock-js/issues/299

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,7 +434,11 @@ expect(res.statusCode).toBe(200);
 
 ## Runner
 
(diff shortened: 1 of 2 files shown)


carolstran:
Added a little section about how to change the runner default, but now I'm wondering if that section should move below the Jest section (since this one is more necessary for usage) 🤔  wdyt?

PR #393, opened 2020-02-27
Sets up basic structure for unmock-xmlhttprequest

mikesol:
This is just a sketch at this point for `xmlhttprequest` support. Not ready for review.


mikesol:
Ready for review, passes tests.

packages/unmock-xmlhttprequest/src/__tests__/response-function.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        ok: true,
+      }),
+    };
+    sendResponse(res);
+  } catch (err) {
+    emitError(err);
+  }
+};
+
+replaceOpenAndReturnOriginal(respondOk);
```
ksaaskil: I think this would persist to other tests if running in band? Would it be safer to switch it off at the end of the test?

packages/unmock-xmlhttprequest/src/__tests__/serialize.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  it("should serialize method correctly", () => {
+    const req: ISerializedRequest = serialize(
+      "https://example.com",
+      "DELETE",
+      {},
+    );
+    expect(req).toHaveProperty("method", "delete");
+  });
+
+  it.skip("should serialize a plain request URL", () => {
```
ksaaskil: This doesn't work yet?

PR #402, opened 2020-03-17
Add unmock.mock to replace unmock.nock.

ksaaskil:
- Adds `unmock.mock` as the preferred equivalent of `unmock.nock`
- Keeps `unmock.nock` for backwards compatibility

Why the change? Now that we're targeting environments beyond Node.js, users cannot be expected to be familiar with [nock](https://github.com/nock/nock). I also think it's better to make it clear that unmock-js is a standalone and independent library not designed as "nock alternative".

What do you think? @mikesol @carolstran

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ const { runner, transform, u } = unmock;
 const { withCodes } = transform;
 
 unmock
-  .nock("https://zodiac.com", "zodiac")
+  .mock("https://zodiac.com", "zodiac")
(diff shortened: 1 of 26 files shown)


mikesol:
Super cool. Allows us to do stuff like:

```typescript
import unmock as un, { u } from 'unmock';
un.mock("https://api.petstore.io").get("/pets").reply(200, u.array({id: u.integer()}));
```

